PR #3, opened 2017-10-31
Add React adapter to Fractal instance

jbalboni:
This adds a React adapter to the fractal instance. I've added the Modal component as an example, though it doesn't look very nice, since we aren't pulling in any vets styles.

I've also moved this over to yarn.

Connects to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5610


jbalboni:
Made a change to add prop docs:
screen shot 2017-11-01 at 4 33 16 pm

theme-overrides/partials/browser/panel-props.nunj
```diff
@@ -0,0 +1,15 @@
+<div class="Browser-panel Browser-props" data-role="tab-panel" id="browser-{{ entity.id }}-panel-props">
```
jbalboni: This is very simplistic and will need updated to fully express all the possible prop types.

ncksllvn:
I for some reason couldn't get the modal demo to open on my `localhost`but other than that it all worked great (output said that the `onClose` prop is required but missing, so maybe that?)

PR #5, opened 2017-11-03
Export only react components

annekainicUSDS:
Connected to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/5769
(^^for some reason it's not letting me connect below using the Zenhub GH extension)

This PR:
1. Runs through original React component files and transforms them using Babel, changing it from jsx to regular js
2. It puts this transformed file of the same name in a directory under `dist/`
3. Creates a npm command to run this script
4. Sets the package `"main"` prop to that directory of transformed React components (but don't think this actually does anything until we publish the npm package)


src/components/modal/Modal.js
```diff
@@ -100,9 +100,6 @@ Modal.propTypes = {
   onClose: PropTypes.func.isRequired,
   title: PropTypes.string,
   visible: PropTypes.bool.isRequired,
-  test: PropTypes.arrayOf(PropTypes.shape({
```
annekainicUSDS: @jbalboni I assumed this was left in by mistake, but I could be wrong

package.json
```diff
@@ -15,6 +23,7 @@
   "devDependencies": {
     "babel-core": "^6.26.0",
     "babel-preset-env": "^1.6.1",
+    "babel-preset-es2015": "^6.24.1",
```
jbalboni: Is this necessary? It should be covered by preset-env and stage-2.

annekainicUSDS: Good question, I thought I had to revert back to using es2015 instead of stage-2, but maybe that's covered in the change to the .babelrc file and doesn't need to be separately installed here?

PR #7, opened 2017-11-07
Set up unit and a11y testing

jbalboni:
This sets up unit testing for our components. There are a few decisions made that are worth pointing out and discussing:

1. Using Karma
2. Putting test in same directory as component
3. a11y testing in unit tests, not e2e tests
4. Using Enzyme

I chose 1 because it's easier to debug and we can run tests in a real browser, which makes it possible to do 3. I think that's an advantage for these types of visual components, where we care primarily about how they work in a browser and are not testing business or data transformation logic as much.

I find 2 to be easier in general, but it's a preference and we can always change it later.

The downside of using Karma is that it is not always super reliable and we also have to use Webpack to compile our code before running it in headless Chrome, which isn't super fast. But I think this repo will stay small enough that those are major problems.

For 4, it seemed like we might be headed that way, so I started using it here.

The two new tasks are `yarn test` and `yarn test-debug`.


jbalboni:
@cehsu I'm not too worried about those right now, we probably want to turn off prop type warnings globally in tests.

I think we'll hook up CI eventually, but I don't think we need it yet.

PR #8, opened 2017-11-09
Add segmented progressbar

annekainicUSDS:
`SegmentedProgressBar` is added to design-system, and importing it for use in vets-website works great. Encountered a couple of issues in getting this to work:

1. Not confident that the updates to components.bundle.js are correct. Tried deleting the file entirely and restarting the server, which should re-render it, but there's an error recorded in components.bundle.js that I don't think should be.
2. I had a lot of trouble getting the styles to show up. I realized this was because the npm package node-sass wasn't installed. What eventually worked for me was running `npm install node-sass` directly. Not sure why I had to do this, as node-sass is specified in the devDependencies of package.json.


sass/site.scss
```diff
@@ -2,3 +2,4 @@
 
 @import "style";
 @import "modules/m-modal";
+@import "modules/m-progress-bar";
```
annekainicUSDS: Did I include this in the right spot? Feels like there's somewhere better to import this, but followed what was done for `Modal`.

jbalboni: Yeah, my intent was that this file would be used to import all the sass for all the components.

dist/styles.bundle.js
```diff
@@ -63,15 +63,15 @@
...
...er-runner/lib/loadLoader.js:36:3)\n    at iteratePitchingLoaders (/Users/jbalboni/adhoc/design-system/node_modules/loader-runner/lib/LoaderRunner.js:169:2)\n    at runLoaders (/Users/jbalboni/adhoc/design-system/node_modules/loader-runner/lib/LoaderRunner.js:362:2)\n    at NormalModule.doBuild (/Users/jbalboni/adhoc/design-system/node_modules/webpack/lib/NormalModule.js:179:3)\n    at NormalModule.build (/Users/jbalboni/adhoc/design-system/node_modules/webpack/lib/NormalModule.js:268:15)\n    at Compilation.buildModule (/Users/jbalboni/adhoc/design-system/node_modules/webpack/lib/Compilation.js:146:10)\n    at moduleFactory.create (/Users/jbalboni/adhoc/design-system/node_modules/webpack/lib/Compilation.js:433:9)\n    at factory (/Users/jbalboni/adhoc/design-system/node_modules/webpack/lib/NormalModuleFactory.js:253:5)\n    at applyPluginsAsyncWaterfall (/Users/jbalboni/adhoc/design-system/node_modules/webpack/lib/NormalModuleFactory.js:99:14)\n    at /Users/jbalboni/adhoc/design-system/node_modules/tapable/lib/Tapable.js:268:11\n    at NormalModuleFactory.params.normalModuleFactory.plugin (/Users/jbalboni/adhoc/design-system/node_modules/webpack/lib/CompatibilityPlugin.js:52:5)");
```
jbalboni: Yeah, this shouldn't have been checked in. I'm assuming that was me, but I'm not sure how it happened.

PR #10, opened 2017-11-14
Add unit tests and comments progress bar

annekainicUSDS:


lib/testing/helpers.js
```diff
@@ -1,17 +1,19 @@
 import { mount } from 'enzyme';
+import _ from 'lodash';
 
 export function axeCheck(component) {
   let div = document.getElementById('axeContainer');
+  let uniqueId = _.uniqueId('container-');
   if (!div) {
     div = document.createElement('div');
-    div.setAttribute('id', 'container');
+    div.setAttribute('id', uniqueId);
```
jbalboni: I think this code will work without the umounting and unique id additions, if we update the id on this line to match the one on line 5. The intent is that we always reuse that div, and only create it if it's the first time and that hasn't been done yet.

src/components/segmented_progress_bar/SegmentedProgressBar.jsx
```diff
@@ -19,6 +19,12 @@ export default function SegmentedProgressBar({ current, total }) {
 }
 
 SegmentedProgressBar.propTypes = {
+  /**
+   * The urrent segment in progress
```
jbalboni: Typo

annekainicUSDS: ah thanks!

PR #11, opened 2017-11-20
Add eslint config

jbalboni:
We may want to share this in the future with vets-website, rather than copying it over, but I think this is ok for now.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -85,12 +85,10 @@
     'no-multi-spaces': 0,
     'no-undef-init': 0,
     'padded-blocks': 0,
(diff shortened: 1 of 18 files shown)


.eslintrc
```diff
@@ -0,0 +1,98 @@
+{
```
jbalboni: This is copied from vets-website, except that I turned on the two a11y rules we turned off.

cehsu:
Is using something like eslint-plugin-compat something of potential interest, perhaps in the future?

jbalboni:
@cehsu Maybe, but I think we try to polyfill those things, so I'm not sure how useful it would be.

PR #14, opened 2017-12-20
Migrate errorable checkbox and tooltip components

cehsu:
These changes move the errorable checkbox and tooltip components into the design system, as well as add a few additional tests, and make some accessibility related updates ([reference used](https://w3c.github.io/aria-practices/examples/checkbox/checkbox-1/checkbox-1.html)), and refactor to make use of [generic props](https://github.com/department-of-veterans-affairs/design-system/issues/19) including:

Tooltip
- toggling onBlur and onFocus
- requiring non-negative digit tabIndex

ErrorableCheckbox
- including aria-labelledby
- replace onValueChange and remove handleChange with onChange that targets e.target.checked

Including aria-labelledby revealed that the flat structure of the checkbox was actually against [best practice](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-for.md), and the updated, nested structure requires some additional styling work. Issue created for this work [here](https://github.com/department-of-veterans-affairs/design-system/issues/18).


src/components/errorable_checkbox/ErrorableCheckbox.jsx
```diff
@@ -0,0 +1,105 @@
...
+    return (
+      <div className={className}>
+        <label
+          className={this.props.errorMessage ? 'usa-input-error-label' : undefined}
+          id={labelId}
+          name={`${this.props.name}-label`}
+          htmlFor={this.inputId}>
+          <input
+            autoComplete="false"
+            aria-checked={!!this.props.checked}
```
jbalboni: Not sure if this was your change, but it shouldn't be necessary on a checkbox, I think?

src/components/errorable_checkbox/ErrorableCheckbox.jsx
```diff
@@ -0,0 +1,105 @@
...
+          {requiredSpan}
+        </label>
+        {errorSpan}
+        {toolTip}
+      </div>
+    );
+  }
+}
+
+ErrorableCheckbox.propTypes = {
```
jbalboni: You should add some comments here like on other components for the autogenerated documentation.

src/components/errorable_checkbox/ErrorableCheckbox.jsx
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    // Calculate required.
+    let requiredSpan = undefined;
+    if (this.props.required) {
+      requiredSpan = <span className="form-required-span">*</span>;
+    }
+
+    let className = `form-checkbox${this.props.errorMessage ? ' usa-input-error' : ''}`;
+    if (!_.isUndefined(this.props.className)) {
```
jbalboni: I think we could remove the use of lodash here.

src/components/errorable_checkbox/errorable_checkbox.config.yml
```diff
@@ -0,0 +1,8 @@
+label: ErrorableCheckbox
+status: wip
+
+context:
+  componentSourcePath: './ErrorableCheckbox.jsx'
+  id: default
+  title: 'ErrorableCheckbox'
+  content: 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aliquam id felis pulvinar ligula ultricies sollicitudin eget nec dui. Cras augue velit, pellentesque sit amet nisl ut, tristique suscipit sem. Cras sollicitudin auctor mattis.'
```
jbalboni: We shouldn't need this content data.

src/components/errorable_checkbox/errorable_checkbox.njk
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import ErrorableCheckbox from './ErrorableCheckbox';
+
+export default function ErrorableCheckboxExample(props) {
+  return (
+          <ErrorableCheckbox
+            errorMessage="error message"
```
jbalboni: Might be better to read this from `props`, so that you can set the message in the config.yml file.

src/components/errorable_checkbox/errorable_checkbox.njk
```diff
@@ -0,0 +1,12 @@
...
+            required
+            label="Errorable Checkbox"
```
jbalboni: Same as above.

src/components/errorable_checkbox/errorable_checkbox.njk
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import ErrorableCheckbox from './ErrorableCheckbox';
+
+export default function ErrorableCheckboxExample(props) {
+  return (
+          <ErrorableCheckbox
+            errorMessage="error message"
+            required
```
jbalboni: We could make a Fractal variant for the non-required version, too.

src/components/tooltip/Tooltip.jsx
```diff
@@ -0,0 +1,70 @@
...
+          <button tabIndex={this.props.tabIndex} onClick={this.handleClose}>Close</button>
+        </div>
+      </div>
+    );
+  }
+}
+
+ToolTip.propTypes = {
+  /* eslint-disable consistent-return */
+  tabIndex(props, propName, componentName) {
```
jbalboni: I'm not sure I understand what's happening here. What didn't work about the number prop type from vets-website version?

src/components/tooltip/Tooltip.unit.spec.jsx
```diff
@@ -0,0 +1,18 @@
...
+import Tooltip from './Tooltip.jsx';
+
+describe('<Tooltip/>', () => {
+  it('should render', () => {
+    const tree = shallow(<Tooltip toolTipText="test" tabIndex={0}/>);
+    expect(tree.text()).to.contain('test');
+  });
+
+  it('should pass aXe check', () => {
+    const check = axeCheck(<Tooltip toolTipText="test" tabIndex={0}/>);
```
jbalboni: Can we test the open and closed state?

src/components/tooltip/tooltip.njk
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import Tooltip from './Tooltip';
+
+export default function TooltipExample(props) {
+  return (
+          <Tooltip
+            toolTipText="Tooltip text"
```
jbalboni: Can use the `content` property in `props` to set this.

src/components/errorable_checkbox/ErrorableCheckbox.jsx
cehsu: Good idea.

src/components/errorable_checkbox/ErrorableCheckbox.jsx
cehsu: Good point.

src/components/errorable_checkbox/ErrorableCheckbox.jsx
cehsu: Good idea.

src/components/errorable_checkbox/errorable_checkbox.config.yml
cehsu: Good catch.

src/components/errorable_checkbox/errorable_checkbox.njk
cehsu: Good idea.

src/components/errorable_checkbox/errorable_checkbox.njk
cehsu: Sounds good.

src/components/errorable_checkbox/errorable_checkbox.njk
cehsu: Good idea.

src/components/tooltip/Tooltip.jsx
cehsu: The idea here is to require non-negative tabIndex (for accessibility). Is there a better way to go about this?

src/components/tooltip/Tooltip.unit.spec.jsx
cehsu: Good idea.

src/components/tooltip/tooltip.njk
cehsu: Thanks for the recommendation.

src/components/errorable_checkbox/ErrorableCheckbox.jsx
```diff
@@ -0,0 +1,122 @@
...
+    this.handleChange = this.handleChange.bind(this);
+  }
+
+  componentWillMount() {
+    this.inputId = _.uniqueId('errorable-checkbox-');
+  }
+
+  handleChange(domEvent) {
+    this.props.onValueChange(domEvent.target.checked);
+  }
```
ncksllvn: Just an idea and not super relevant to this PR, but have we considered using component properties that match the wrapped form element? That way the developer doesn't have to learn a new API/look up documentation and we could also do something like `<input {...this.props/>` so we don't have to account for each HTML attribute.

ncksllvn:
Do we need to add syntax highlighting for JSX?

image

cehsu:
Thanks @ncksllvn do you mind creating tickets for those ideas, they're great ones.

src/components/errorable_checkbox/ErrorableCheckbox.jsx
jbalboni: That's definitely better, I think these components are just old and haven't been re-visited in a while.

cehsu: Thanks @ncksllvn, I've replaced the name and checked assignment with {...this.props}. As for your proposal to use component properties that match the wrapped form component, this is an interesting idea, could you please clarify which additional properties (beyond name and checked) we should consider doing this with? Or are you suggesting that all required properties be assigned in the encompassing ErrorableCheckbox perhaps as a nested object:

```
const props = {
  onValueChange: {() => {}},
  label: 'Checkbox',
  required: true,
  inputProps: {
   ...inputProps
  },
  labelProps: {
   ...labelProps
  }
}
```

Which would be used in the child components like so `<input {...this.props.inputProps}/>`?
If so, is this preferable to relying on propTypes and documentation/comments to convey this information?

jbalboni: I think Nick was saying that some of the names we use for props could more closely match the props available on `input`. Like instead of having a prop called `onValueChange`, calling it `onChange` instead, so it matches the prop you would use if you were using `input` directly.

That involves changing the interface of the component, though, so it could take a while and would probably be done in its own PR.

jbalboni:
@ncksllvn The syntax highlighting issue is definitely annoying. Fractal won't let you specify different template languages for different features, so some of the jsx files have to have an `njk` extension (that or we have to convert a bunch of templates to jsx).

src/components/errorable_checkbox/ErrorableCheckbox.jsx
cehsu: Thanks for clarifying @jbalboni, using generic props is a great idea, thanks again @ncksllvn. Ticket created for this [here](https://github.com/department-of-veterans-affairs/design-system/issues/16).

ncksllvn: Yeah that's exactly what I meant - thank you @cehsu, thanks @jbalboni!

PR #23, opened 2018-01-03
Remove build assets, add site publishing command

jbalboni:
This removes all the built assets and changes the commands for running things so that we can publish the design system site more easily.

I did not do the versioned directories thing, but I think that would be fairly straightforward.

The important changes are that we now have two commands to run:

`npm run start` (for building locally)
`npm run site` (for updating the gh io site)

Those need to be run instead of fractal commands now, because we have custom commands to support the site building.


jbalboni:
@ncksllvn Yeah, but it shouldn't be too difficult.

PR #30, opened 2018-01-15
begin migrating AlertBox

phiden:
~@cehsu @jbalboni @annekainicUSDS Hello all -- starting trying to migrate this component and getting 'AlertBox not defined.' No idea why -- any of you have time to help me figure it out? Please & thanks!~

update: it was a typo; now trying to figure out how to make it render.


src/components/05-alerts/AlertBox/AlertBox.unit.spec.jsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import SkinDeep from 'skin-deep';
```
jbalboni: These tests need some updating. In this repo we're switching everything to use Enzyme. You should be able to look at other tests in here and see how the `shallow` function is used and replace `SkinDeep.shallowRender` with that.

Also, we want to have some tests that use Enzyme's `mount` and run the aXe check (there should be examples of that, too, in this repo).

src/components/05-alerts/AlertBox/AlertBox.jsx
```diff
@@ -0,0 +1,89 @@
...
+        </div>
+        {closeButton}
+        <div className="cf"></div>
+      </div>
+    );
+  }
+
+}
+
+AlertBox.propTypes = {
```
jbalboni: We should add some documentation in here to explain what each of these properties do.

phiden: will do

src/components/05-alerts/AlertBox/AlertBox.unit.spec.jsx
phiden: Ok, I'll work on this -- thanks!

phiden:
@jbalboni I got about halfway there, I think -- swapped in enzyme, and am running an axecheck -- but could you give me more context around what you mean by using `mount` and what other tests you would expect to see, please?

jbalboni:
@phiden You can disregard what I said about mount, I forgot how I had the axeCheck set up. What you did is what I was looking for. As for other tests, I think we want to run the axeCheck on different visual states of the component: visible and not visible (since the "not visible" state actually renders a div), with and without a close button, and with and without a headline.

phiden:
@jbalboni added some tests -- please let me know if I did it wrong!

PR #31, opened 2018-01-22
add errorable radio button and its dependencies

rianfowler:
screen shot 2018-01-22 at 09 10 31
screen shot 2018-01-22 at 09 10 44
Pushing what I have so far. Both are tagged WIP in the design system.

I added ExpandingGroup and a helper from the website to fulfill ErrorableRadioButtons dependencies. 

I was unable to get any of the js dependent interactions to work on the local version but I think this is working as intended. LMK if I need to do something differently.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "react-docgen": "^2.19.0",
     "react-dom": "^15.6.2",
     "react-test-renderer": "^15.5.0",
+    "react-transition-group": "^1.1.2",
     "resolve-url-loader": "^2.2.0",
     "sass-loader": "^6.0.6",
     "style-loader": "^0.19.0",
--- a/src/components/errorable_radio_buttons/ErrorableRadioButtons.jsx
+++ b/src/components/errorable_radio_buttons/ErrorableRadioButtons.jsx
@@ -1,0 +1,198 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import _ from 'lodash';
+import classNames from 'classnames';
+
+import ToolTip from '../tooltip/Tooltip';
+import ExpandingGroup from '../expanding_group/ExpandingGroup';
+
+import { makeField } from '../../model/fields.js';
+
+/**
+ * A radio button group with a label.
+ *
+ * Validation has the following props.
+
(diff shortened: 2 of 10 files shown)


src/model/fields.js
```diff
@@ -0,0 +1,28 @@
+import _ from 'lodash';
+
+/**
+ * Represents an input field value.
+ *
+ * Input fields need to have a `dirty` state that represents whether or not a user has touched it.
+ * Without this state, it is extremely hard (impossible?) to write UI with required fields where
+ * the initial empty state does not get marked as a distracting error.
+ */
+export function makeField(value, optionalDirty) {
```
rianfowler: ErrorableRadioButtons uses this function. I added the whole fields.js file and put it in a folder structure that mostly matches the websites repo.

src/components/errorable_radio_buttons/errorable_radio_buttons.config.yml
```diff
@@ -0,0 +1,18 @@
+label: ErrorableRadioButtons
+status: wip
+
+context:
```
rianfowler: I tried to demonstrate all of the propTypes that visually render something.

src/components/errorable_radio_buttons/ErrorableRadioButtons.jsx
```diff
@@ -0,0 +1,198 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import _ from 'lodash';
+import classNames from 'classnames';
+
+import ToolTip from '../tooltip/Tooltip';
+import ExpandingGroup from '../expanding_group/ExpandingGroup';
+
+import { makeField } from '../../model/fields.js';
```
rianfowler: Localized some of the imports. Otherwise the file is unchanged.

package.json
```diff
@@ -60,6 +60,7 @@
     "react-docgen": "^2.19.0",
     "react-dom": "^15.6.2",
     "react-test-renderer": "^15.5.0",
+    "react-transition-group": "^1.1.2",
```
rianfowler: ExpandingGroup dependency.

PR #33, opened 2018-01-22
6506 move errorable radio buttons

rianfowler:


createWebpackBundle.js
```diff
@@ -137,10 +137,21 @@ function createWebpackBundle(logger, fractalComponents, watch = true) {
...
       if (err || stats.hasErrors()) {
         const info = stats.toJson();
         logger.error(info.errors);
+      } else {
+        if (!buildHashes.includes(stats.hash)) {
+          buildHashes.push(stats.hash);
+          logger.log(stats.toString('minimal'));
+
+          if (buildHashes.length > 1) {
+            logger.log('\x1b[36mBuild complete');
```
rianfowler: screen shot 2018-01-22 at 14 26 33
This tripped me up so I added some additional logging to the webpack build process. Should be clear now when you can load the site and see the components.

src/components/errorable_radio_buttons/errorable_radio_buttons.njk
```diff
@@ -0,0 +1,32 @@
...
+export default function ErrorableRadioButtonsExample(props) {
+  return (
+    <StateWrapper
+      {...props}
+      >
+        <ErrorableRadioButtons
+          onValueChange={()=>{}}
+          {...props}
+         />
+      </StateWrapper>
```
rianfowler: Wrapped this in something I could trigger a rerender with.

src/components/expanding_group/expanding_group.njk
```diff
@@ -0,0 +1,38 @@
...
+      </div>
+    );
+  }
+}
+
+export default function ExpandingGroupExample(props) {
+  return (
+    <Toggler
+      {...props}
+    />
```
rianfowler: The radio buttons had this as a dependency so I added it with a toggling button as well.
screen shot 2018-01-22 at 15 09 46

src/model/fields.js
```diff
@@ -0,0 +1,28 @@
...
+ * Represents an input field value.
+ *
+ * Input fields need to have a `dirty` state that represents whether or not a user has touched it.
+ * Without this state, it is extremely hard (impossible?) to write UI with required fields where
+ * the initial empty state does not get marked as a distracting error.
+ */
+export function makeField(value, optionalDirty) {
+  const dirty = optionalDirty === undefined ? false : optionalDirty;
+  return { value, dirty };
+}
```
rianfowler: ErrorableRadioButtons uses this function. I added the whole folder/file to somewhat mirror what's in the website repo.

src/components/expanding_group/expanding_group.config.yml
```diff
@@ -0,0 +1,7 @@
+label: ExpandingGroup
+status: wip
+notes: Expects two child components.
+
+context:
+  componentSourcePath: './ExpandingGroup.jsx'
+  open: true
```
rianfowler: For some reason I couldn't get this to translate a falsey value into a native js type. `True`, `true` work but `false`, `False` become `null`.

README.md
```diff
@@ -166,7 +166,6 @@ Deployment steps assume you've followed the getting started section at the top o
 
 GitHub will automatically try to render any Markdown files using Jekyll. If you get any Jekyll build errors from Github (via email, usually) after pushing the site, you can run the site locally to address any errors and bypass looping through the above. Instructions here assume you already have Bundler installed (see [instructions](https://help.github.com/articles/setting-up-your-github-pages-site-locally-with-jekyll/) if you don't know what that is).
 
-* Run `fractal build`
-* Run `bundle exec jekyll serve -s build`
+* Run `npm run start`
 
-Add `-t` to the above if you want to get tracebacks from any errors.
+If the `fractal watch` task fails, remove the `dist` directory and try running `npm run start` again.
```
rianfowler: Updated readme instructions.

package.json
```diff
@@ -60,6 +60,7 @@
     "react-docgen": "^2.19.0",
     "react-dom": "^15.6.2",
     "react-test-renderer": "^15.5.0",
+    "react-transition-group": "^1.1.2",
```
jbalboni: Can we add this to the regular dependencies instead?

src/components/errorable_radio_buttons/ErrorableRadioButtons.unit.spec.jsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import ReactTestUtils from 'react-dom/test-utils';
+import SkinDeep from 'skin-deep';
```
jbalboni: We've been converting tests to Enzyme, so you'll want do that here, as well. The CI isn't set up to run them, so you'll have to run them locally and make sure they pass there.

We also add aXe tests for the different visual states of a component, which you can see examples of on other React components in here.

src/components/errorable_radio_buttons/ErrorableRadioButtons.jsx
```diff
@@ -0,0 +1,198 @@
...
+        </legend>
+        {errorSpan}
+        {optionElements}
+        {toolTip}
+      </fieldset>
+    );
+  }
+}
+
+ErrorableRadioButtons.propTypes = {
```
jbalboni: We should add some comments in here as documentation, so it shows up on the Props tab in Fractal.

package.json
rianfowler: Yup- do you know why it's in the `devDependencies` in the website repo?

src/components/errorable_radio_buttons/ErrorableRadioButtons.jsx
```diff
@@ -0,0 +1,240 @@
...
+        additional: PropTypes.oneOfType([
+          PropTypes.string,
+          PropTypes.element
+        ])
+      })
+    ])).isRequired,
+  /**
+   * value object for selected field <br/>
+   * value: string value that matches radio button value </br>
+   * dirty: indicates if form is dirty; should be true after any user input
```
rianfowler: Maybe there's a better way than using `<br/>` here. Do we know what's generating the documentation? Maybe it has some better formatting options.

src/components/errorable_radio_buttons/ErrorableRadioButtons.unit.spec.jsx
```diff
@@ -0,0 +1,62 @@
+import React from 'react';
```
rianfowler: First time using aXe and enzyme. Please review carefully.

src/components/errorable_radio_buttons/ErrorableRadioButtons.jsx
```diff
@@ -0,0 +1,240 @@
...
+   *
+   */
+  options: PropTypes.arrayOf(
+    PropTypes.oneOfType([
+      PropTypes.string,
+      PropTypes.shape({
+        label: PropTypes.oneOfType([
+          PropTypes.string,
+          PropTypes.element,
+        ]).isRequired,
```
rianfowler: `label` and `value` are required for this component to render in an accessibility compliant way.

src/components/errorable_radio_buttons/ErrorableRadioButtons.jsx
```diff
@@ -0,0 +1,240 @@
...
+      const radioButton = (
+        <div key={optionAdditional ? undefined : index} className="form-radio-buttons">
+          <input
+            autoComplete="false"
+            checked={checked}
+            id={`${this.inputId}-${index}`}
+            name={this.props.name}
+            type="radio"
+            value={optionValue}
+            onChange={() => this.handleChange(optionValue)} />
```
rianfowler: Refactored this handler to explicitly pass the `optionValue` from the component closure rather than have the handler extract the value from the DOM event. 

Enzyme doesn't simulate real DOM events, so this was blowing up during the unit test.

rianfowler:
I think I've addressed all of your comments @jbalboni

src/components/errorable_radio_buttons/ErrorableRadioButtons.jsx
jbalboni: Enzyme has two ways of running tests, `shallow` and `mount`. `shallow` doesn't use the DOM at all, but `mount` will render the component into the DOM (jsdom, in our case) and should have real DOM events. I think that might be the better option here.

src/components/errorable_radio_buttons/ErrorableRadioButtons.jsx
jbalboni: It's pretty basic. It's using https://github.com/reactjs/react-docgen to get the info from the props, called here: https://github.com/department-of-veterans-affairs/design-system/blob/master/lib/helpers/generatePropDocs.js, and output here: https://github.com/department-of-veterans-affairs/design-system/blob/master/theme-overrides/partials/browser/panel-props.nunj

src/components/errorable_radio_buttons/ErrorableRadioButtons.unit.spec.jsx
```diff
@@ -0,0 +1,62 @@
...
+    expect(valueChanged.dirty).to.eql(true);
+  });
+
+  it('renders label htmlFor attribute with correct input id attribute value', () => {
+    const wrapper = shallow(<ErrorableRadioButtons label="test" options={nonExpandingOptions} value={makeField('test')} onValueChange={(value) =>  value}/>);
+
+    // gather input id and label for attributes from render component
+    const inputIds = []
+    const labelFors = [];
+    wrapper.find('input').forEach((inputId) => inputIds.push(inputId.prop('id')));
```
jbalboni: This is a minor thing, but I think I would flip your use of forEach and map here.

```js
const inputIds = wrapper.find('input').map((input) => input.prop('id'));
const labelFors = wrapper.find('label').map((labelFor) => labelFor.prop('htmlFor'));

inputIds.forEach((inputId, index) => expect(inputId).to.eql(labelFors[index]));
```

Since in the first case you're transforming a list into another list and at the end you're calling a function with side effects on each item.

src/components/expanding_group/ExpandingGroup.jsx
```diff
@@ -0,0 +1,41 @@
...
+          ? <div key="removable-group" className={additionalClass}>
+            {children[1]}
+          </div>
+          : null}
+      </ReactCSSTransitionGroup>
+    </div>
+  );
+}
+
+ExpandingGroup.propTypes = {
```
jbalboni: I think we should add some prop comment documentation here, too.

PR #34, opened 2018-01-23
6504 move errorable text input

rianfowler:
screen shot 2018-01-23 at 14 14 12


src/components/errorable_text_input/ErrorableTextInput.jsx
```diff
@@ -0,0 +1,170 @@
...
+
+import ToolTip from '../tooltip/ToolTip.jsx';
+
+import { makeField } from '../../model/fields.js';
+
+/**
+ * A form input with a label that can display error messages.
+ *
+ * Props:
+ * `errorMessage` - Error string to display in the component.
```
jbalboni: I wonder if we should remove this comment block. It should mostly be in the prop types now.

PR #35, opened 2018-01-24
Migrate OMBInfo component

phiden:


src/components/OMBInfo/OMBInfo.jsx
```diff
@@ -0,0 +1,60 @@
...
+}
+
+OMBInfo.propTypes = {
+  // respondent burden, length of time usually in minutes
+  resBurden: PropTypes.number,
+
+  // OMB control number / form number
+  ombNumber: PropTypes.string,
+
+  // form expiration date?
```
jbalboni: Yeah, that's what it is.

PR #36, opened 2018-01-25
Migrate progress button

phiden:


package.json
```diff
@@ -61,6 +60,7 @@
     "react-docgen": "^2.19.0",
     "react-dom": "^15.6.2",
     "react-test-renderer": "^15.5.0",
+    "react-transition-group": "1.x",
```
jbalboni: Do you know why this got moved?

src/components/07-form/controls/ProgressButton/ProgressButton.jsx
```diff
@@ -0,0 +1,60 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import _ from 'lodash';
+
+/**
+ * A component for the continue button to navigate through panels of questions.
+ *
+ * Required props
```
jbalboni: We could probably remove these comments now, given the prop type comments.

src/components/07-form/controls/ProgressButton/ProgressButton.unit.spec.jsx
```diff
@@ -0,0 +1,41 @@
...
+          disabled={false}
+          onButtonClick={() => { resolve(true); }}/>
+      );
+    });
+
+    const button = ReactTestUtils.findRenderedDOMComponentWithTag(progressButton, 'button');
+    ReactTestUtils.Simulate.click(button);
+
+    return expect(updatePromise).to.eventually.eql(true);
+  });
```
jbalboni: Could add some axeCheck tests.

src/components/07-form/controls/ProgressButton/ProgressButton.unit.spec.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+import ProgressButton from './ProgressButton.jsx';
+
+chai.use(chaiAsPromised);
+
+//    const tree = shallow(<Modal
+// id="modal" title="Modal title" visible onClose={() => {//}}>Modal contents</Modal>);
+
+describe('<ProgressButton>', () => {
+  it('has sane looking features', () => {
```
jbalboni: This has always kinda bothered me as a name/description for a test. Maybe `should render with button text` instead?

package.json
phiden: Yes -- the console was reporting missing modules when I was trying to get tests to pass. I tried a few permutations and this was the one that worked, but if it's an issue can always move it around again!

src/components/07-form/controls/ProgressButton/ProgressButton.unit.spec.jsx
phiden: LOL will make the change

package.json
jbalboni: It should already be included in master, so I'd like to revert this and the lock file change in this PR if we could. react-transition-group should be in the dependencies object vs devDependencies so that it gets installed when someone installs the npm module we're building.

PR #37, opened 2018-01-26
WIP - npm-modulize components 

rianfowler:
Looking for feedback on these. 

- I updated the export script to target all `.jsx` files. Let me know if there are other factors to consider when determining what to publish (e.g. only including `.jsx` files that have `.njk` or `.yml` sibling files). 
- I'm going to use [yalc](https://github.com/whitecolor/yalc) to test publishing and importing these next week. As long as no one has problems with global npm modules,  I'm planning on including some instructions in the README on how to use this.
- I think we should add a watch-build-publish command in the design system so developers can edit component in the design system and see it update in the locally running vets-website. Not sure if this would work with `yalc` yet.
- How do you want to handle semantic versioning? We can do anything from documenting in detail on how to use `npm update` and `npm publish` to having a single command that just does a major revision every time the modules are published. 
- This may need more reviewers.


package.json
```diff
@@ -1,6 +1,7 @@
 {
-  "name": "vets.gov-design-system",
-  "version": "0.1.0",
+  "name": "vets.gov-jean-pants",
+  "version": "1.0.0",
+  "license": "MIT",
```
rianfowler: Since this is being publish to npm, do we want to add a license?

package.json
```diff
@@ -1,6 +1,7 @@
 {
-  "name": "vets.gov-design-system",
-  "version": "0.1.0",
+  "name": "vets.gov-jean-pants",
```
rianfowler: 👖

cehsu: Is this meant to just be "jean-pants"?

package.json
cehsu: Good question. "MIT" seems consistent with the roadrunner repository's license, but I'll wait for others to chime in.

package.json
rianfowler: I saw that other packages namespaced to @department-of-veterans-affairs. Not sure how we want to publish this.

package.json
jbalboni: Let's use the same one that vets-website uses.

package.json
jbalboni: Yeah, it should be in the department-of-veterans-affairs org. So the name would be `@department-of-veterans-affairs/jean-pants`

jbalboni:
@rianfowler Transforming jsx files in the components directory I think is fine. We would also need to transform and copy the helper files those components use.

package.json
rianfowler: Sounds good. Will add `CC0-1.0` to the license field as per https://spdx.org/licenses/

jbalboni:
For points 2 and 3, that sounds good to me. I haven't used yalc, but seems helpful.

For point 4, I think we can document how to use those commands and let people choose the right version.

We'll definitely want more reviewers and maybe to go over this in a FE or eng meeting. But we can get things ironed out a bit more before that, if that's easier.

rianfowler:
I'll share this PR in the FE channel after I make some updates and have a draft of `yalc` instructions that I've verified.

rianfowler:
Ah yeah- forgot about the helper files. We may have to use webpack, then.

jbalboni:
Webpack will build things into a single file, which I don't think we want to do. I think we can make a decision to put helper files in one place and transform that folder along with the components. This is what rjsf does, basically: https://github.com/mozilla-services/react-jsonschema-form/blob/master/package.json#L7

rianfowler:
Alrighty- I think I can make that work.

README.md
```diff
@@ -170,3 +170,25 @@ GitHub will automatically try to render any Markdown files using Jekyll. If you
...
+
+[yalc](https://github.com/whitecolor/yalc) is recommended for testing module publication- see installation instructions.
+
+* With yalc installed, run `npm run test-publish`
+
+Copies `package.json` to the `dist/jean-pants` directory and publishes the exported components to the local yalc directory as a `@department-of-veteran-affairs/jean-pants` module.
+
+* In the `vets-website` project, run `yalc add @department-of-veteran-affairs/jean-pants` or `yalc update`
+
+Installs `jean-pants` to `node_modules` making it available for importing and testing locally.
```
rianfowler: I tried this on the feedback form in the footer on the facilities locator with the `ErrorableTextInput` component.
 
screen shot 2018-01-30 at 11 02 18

The import statement is kind of gnarly right now- which should be fine considering their low volatility. Initial testing looked good.

karma.conf.js
```diff
@@ -38,6 +38,9 @@ module.exports = function karmaConfig(config) {
           },
         ]
       },
+      resolve: {
+        extensions: ['.js', '.jsx']
```
rianfowler: Babel exports `.jsx` as `.js` with the current settings. This is desirable for a single import syntax across the design system and vets website. 

Current Karma settings were requiring a `.jsx` import in the design system which doesn't work across the projects with the current babel settings. 

I found / tried this setting and it seems to work - unit tests pass and design system works. 

iu-2

fractal.js
```diff
@@ -7,7 +7,7 @@ const createWebpackBundle = require('./createWebpackBundle');
 
 const context = {
   'package': {
-    name: pkg.name,
+    name: pkg.name.replace('@', ''), // fractal interprets @ as component references when its injected into their contexts
```
rianfowler: Adding the namespace to the module name caused fractal to spam warnings on startup. I believe Fractal was interpreting the `@` on the project name as some sort of context reference that was it was unable to resolve in the project. Removing the `@` from the name string eliminated the warnings.

package.json
```diff
@@ -1,17 +1,18 @@
 {
-  "name": "vets.gov-design-system",
-  "version": "0.1.0",
+  "name": "@department-of-veteran-affairs/jean-pants",
+  "version": "1.0.0",
+  "license": "CC0-1.0",
```
rianfowler: Creative Commons Zero 1.0 Universal

package.json
```diff
@@ -1,17 +1,18 @@
...
   "repository": {
     "type": "git",
     "url": "git+https://github.com/department-of-veterans-affairs/design-system.git"
   },
-  "main": "dist/js/components",
   "scripts": {
+    "export-components": "rimraf dist/jean-pants && babel src/components --out-dir dist/jean-pants/components --ignore .spec.jsx,.njk,.yml && babel src/helpers --out-dir dist/jean-pants/helpers",
```
rianfowler: `README.md` details this script. 

`npm publish` preserves the folder structure from where it's ran so this sets up `jean-pants` to be the root of the module. 

Took some inspiration from `react-jsonschema-form` per @jbalboni 's suggestion. Got rid of the `export-components` script in favor of using a simpler `babel-cli` command. The trade off is less flexibility in the project / module structure. `rimraf` does `rm -rf`. 

Will need to add additional `.gitignore` entries to the `vets-website` for the `yalc` stuff.

screen shot 2018-01-30 at 11 02 44

package.json
```diff
@@ -1,17 +1,18 @@
...
     "prestart": "node prebuild.js",
-    "export-components": "node src/js/export-components.js",
     "start": "fractal watch",
-    "test": "karma start --single-run",
     "site": "fractal build-site && gh-pages -d build",
+    "test": "karma start --single-run",
+    "test-publish": "cp package.json dist/jean-pants/package.json && cd dist/jean-pants && yalc publish && cd ../../",
```
rianfowler: `README.md` details this script.

Copies `package.json` to the module's root folder and runs `yalc publish`.

package.json
```diff
@@ -61,7 +64,9 @@
     "react-docgen": "^2.19.0",
     "react-dom": "^15.6.2",
     "react-test-renderer": "^15.5.0",
+    "recursive-readdir": "^2.2.1",
```
rianfowler: TODO: remove unnecessary dependencies.

rianfowler:
I added additional comments and updates @jbalboni @cehsu. Looking for feedback and thoughts on next steps.

karma.conf.js
jbalboni: Ha, I think that's all good.

README.md
jbalboni: Yeah, it'd be nice if we could remove that sub-folder in components/, but that would cause other problems with relative urls. We could probably write a babel transform to fix the urls, but that would get ugly. Maybe worth a separate ticket.

src/components/05-alerts/AlertBox/AlertBox.unit.spec.jsx
```diff
@@ -3,7 +3,7 @@ import { shallow } from 'enzyme';
 import { axeCheck } from '../../../../lib/testing/helpers';
 import { expect } from 'chai';
 
-import AlertBox from './AlertBox.jsx';
+import AlertBox from './AlertBox';
```
cehsu: Curious what needed to be changed to be able to remove the extensions here?

README.md
```diff
@@ -170,3 +170,25 @@ GitHub will automatically try to render any Markdown files using Jekyll. If you
...
+
+2. Test the components
+
+[yalc](https://github.com/whitecolor/yalc) is recommended for testing module publication- see installation instructions.
+
+* With yalc installed, run `npm run test-publish`
+
+Copies `package.json` to the `dist/jean-pants` directory and publishes the exported components to the local yalc directory as a `@department-of-veteran-affairs/jean-pants` module.
+
+* In the `vets-website` project, run `yalc add @department-of-veteran-affairs/jean-pants` or `yalc update`
```
cehsu: Rephrasing this as ", and subsequently `yalc update`" might be a little clearer.

README.md
cehsu: Good point, could we create a class for the React components to make them easier to import? (e.g. [react-primitives](https://github.com/lelandrichardson/react-primitives/blob/master/src/ReactPrimitives.js))

rianfowler: I'm rusty on import / require behavior but was going to see what we could do to simplify this e.g. with an object in an `index.js` file. 

I want to avoid causing webpack to include something that's not explicitly used and I know this approach accomplishes that. 

Maybe the best approach is to make separate npm modules and just namespace them to `@dfa/jean-pants` and publish the helpers as a peer dependency.

src/components/05-alerts/AlertBox/AlertBox.unit.spec.jsx
rianfowler: The update to the karma config did the trick. I don't understand why it works. I think by default when there's no extension on the import statement Karma only resolves `.js` and adding that option expands its search parameters. 

https://github.com/department-of-veterans-affairs/design-system/pull/37/files/874b9d3d9b3e15b86b09259ef545fa3afe979dc1#diff-a2a3b7b0c9c3b4b93b4aebf4e3ec3cfb

README.md
jbalboni: I think the object in index.js approach would result in all components being in a Webpack bundle that used anything from this module, but it wouldn't hurt to see.

I don't like the idea of publishing each component as a module, but maybe helpers as its own module would accomplish the same thing? Also worth exploring (not necessarily in this PR, though).

jbalboni: Alternatively, maybe we should make the folder names match the component name and change the component file names to `index.jsx`. That way you could do `@department-of-veteran-affairs/jean-pants/components/OMBInfo` instead of `@department-of-veteran-affairs/jean-pants/components/OMBInfo/OMBInfo` (for example)

jbalboni:
@rianfowler I tried this out with yalc, everything went pretty smoothly once I got the import statement typed out correctly.

README.md
rianfowler: That'll be easy. I'll see what kind of feedback we get from the FE channel and update the PR.

PR #38, opened 2018-01-29
Migrate errorable select

phiden:


src/components/07-form/controls/ErrorableSelect/ErrorableSelect.jsx
```diff
@@ -0,0 +1,168 @@
...
+import _ from 'lodash';
+
+import ToolTip from '../../../tooltip/tooltip.jsx';
+
+import { makeField } from '../../../../model/fields.js';
+
+/**
+ * A form select with a label that can display error messages.
+ *
+ * Validation has the following props.
```
jbalboni: Can delete this if it's all covered in prop type docs below.

src/components/07-form/controls/ErrorableSelect/ErrorableSelect.jsx
```diff
@@ -0,0 +1,168 @@
...
+      PropTypes.string,
+      PropTypes.shape({
+        label: PropTypes.string,
+        value: PropTypes.number }),
+      PropTypes.shape({
+        label: PropTypes.string,
+        value: PropTypes.string }),
+    ])).isRequired,
+
+  // boolean. Render marker indicating field is required.
```
jbalboni: I don't think we need to say "string" or "boolean" in these, unless it reads better that way to you. The type will show up in Fractal next to this message.

src/components/07-form/controls/ErrorableSelect/ErrorableSelect.jsx
```diff
@@ -0,0 +1,168 @@
...
+    dirty: PropTypes.bool
+  }).isRequired,
+
+  // a function with this prototype: (newValue)
+  onValueChange: PropTypes.func.isRequired,
+
+  // ?
+  additionalClass: PropTypes.string,
+
+  // ?
```
jbalboni: This is the description that shows up for the blank option, when `includeBlankOption` is true. Maybe move this prop type up next to that one?

src/components/07-form/controls/ErrorableSelect/ErrorableSelect.jsx
```diff
@@ -0,0 +1,168 @@
...
+  *   - `dirty`: boolean. Whether a field has been touched by the user.*/
+  value: PropTypes.shape({
+    value: PropTypes.string,
+    dirty: PropTypes.bool
+  }).isRequired,
+
+  // a function with this prototype: (newValue)
+  onValueChange: PropTypes.func.isRequired,
+
+  // ?
```
jbalboni: Additional css class that is added to the select element.

src/components/07-form/controls/ErrorableSelect/ErrorableSelect.jsx
```diff
@@ -0,0 +1,168 @@
...
+  // When defined, indicates select has a validation error.
+  errorMessage: PropTypes.string,
+
+  // String for the select name attribute.
+  name: PropTypes.string,
+
+  // String for the select autocomplete attribute.
+  autocomplete: PropTypes.string,
+
+  //String for the select field label.
```
jbalboni: Not always a string, could be a React element.

src/components/07-form/controls/ErrorableSelect/ErrorableSelect.unit.spec.jsx
```diff
@@ -0,0 +1,120 @@
+import React from 'react';
+import ReactTestUtils from 'react-dom/test-utils';
```
jbalboni: We can replace the use of this with the `mount` function from enzyme.

phiden: @jbalboni I tried to make this change but I definitely don't know what I'm doing. Any resource you can point me to / guidance you can provide would be most appreciated!

phiden:
I have no idea why DS_Store is getting committed -- it's in .gitignore. This is frustrating.

src/components/07-form/controls/ErrorableSelect/ErrorableSelect.unit.spec.jsx
jbalboni: Yeah, it doesn't help that this test is weird and unnecessarily complicated. I would completely replace it with the first test in the file for ErrorableTextInput and tweak that to use ErrorableSelect (I think you would just need to pass in options and use `select` instead of `input` in one spot).

phiden:
THANK YOU @jbalboni -- I'd have been spinning my wheels on that forever! LMK if you prefer a clean PR for this guy or if it looks good to merge. Please & thanks!

PR #39, opened 2018-01-30
migrate loading indicator

phiden:
minus the 800+ files that have already been reviewd


src/components/loading_indicator/LoadingIndicator.jsx
```diff
@@ -0,0 +1,35 @@
...
+          aria-valuetext={message}
+          tabIndex="0">
+        </div>
+        {message}
+      </div>
+    );
+  }
+}
+
+LoadingIndicator.propTypes = {
```
jbalboni: We should add some prop type documentation

PR #40, opened 2018-01-31
Reorg

phiden:
This is a first pass at reorganizing design system into something more logical. There will be more passes as we migrate more components!


rianfowler:
The way the [module export](https://github.com/department-of-veterans-affairs/design-system/pull/37#discussion_r164820374) is being set up, the file structure will determine how components are imported into `vets-website`. 

For example [SegmentedProgressBar](https://github.com/department-of-veterans-affairs/design-system/blob/a2e083a59415391dd0c2db51119b77fff005c017/src/components/07-form/SegmentedProgressBar/SegmentedProgressBar.jsx) will be imported as `@department-of-veteran-affairs/jean-pants/components/07-form/SegmentedProgressBar/SegmentedProgressBar`.

If we update the name of the component of `index.jsx` then we can trim off the last segment of the path for all of its imports: `@department-of-veteran-affairs/jean-pants/components/07-form/SegmentedProgressBar/SegmentedProgressBar`.

jbalboni:
I would like for the imports for vets-website to be as simple as possible, which means having less nesting and more of a flat organization in the components folder. A folder like `07-form` is not something a developer would remember and would take some trial and error to figure out.

But not having that structure makes the Fractal site harder to understand.

I'm not really sure how best to balance these things.

phiden:
@jbalboni is there a naming convention that would be easier to understand for developers that would also support the Fractal structure?

rianfowler:
I'll do some additional research on babel transforms. This has to be a problem that someone else has solved.

phiden:
In the meantime, thoughts on merging this PR?

rianfowler:
Looks good to me. It'll be easier for me to figure out something that works if I'm going off the file structure that works for the design system.

PR #43, opened 2018-02-08
Migrate breadcrumb

phiden:
This is the first attempt to make a breadcrumbs component generic.

### DIFF
--- a/src/components/Breadcrumbs/Breadcrumbs.jsx
+++ b/src/components/Breadcrumbs/Breadcrumbs.jsx
@@ -1,0 +1,42 @@
+import { Link, browserHistory } from 'react-router';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+class Breadcrumbs extends React.Component {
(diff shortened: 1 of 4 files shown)


src/components/Breadcrumbs/Breadcrumbs.jsx
```diff
@@ -0,0 +1,27 @@
...
+import React from 'react';
+
+class Breadcrumbs extends React.Component {
+  render() {
+
+    const crumbs = this.props.crumbs;
+    return (
+      <nav className="va-nav-breadcrumbs">
+        <ul className="row va-nav-breadcrumbs-list columns" role="menubar" aria-label="Primary">
+          {crumbs.map((c) => {
```
rianfowler: When using an iterator like `map` or `reduce` it's a good idea to use full singular and plural names e.g. `crumbs` and `crumb`.

rianfowler:
Looks good!

PR #49, opened 2018-03-08
remove extensions from imports

rianfowler:


package.json
```diff
@@ -77,5 +77,8 @@
     "svg-url-loader": "^2.3.0",
     "url-loader": "^0.6.2",
     "webpack": "2"
+  },
+  "peerDependencies": {
+    "react": "^15.5.4"
```
rianfowler: I'm out of my depth here but I tested this and it seems to work. It's a little weird to me to have to list react as a peer and dev dependency but the build failed otherwise. I also changed the react version to match what we're using on `vets-website`

src/components/OMBInfo/OMBInfo.jsx
```diff
@@ -1,6 +1,6 @@
 import PropTypes from 'prop-types';
 import React from 'react';
-import Modal from '../modal/Modal.jsx';
+import Modal from '../modal/Modal';
```
rianfowler: I forgot to do this earlier but oddly it didn't come up in my testing until today. Rather than try to find a way to transform these paths `.jsx` -> `.js`, I'm removing them all together. This merged [update](https://github.com/department-of-veterans-affairs/design-system/blob/7087bc1eae4d0ddb898ce2d92dc47d018263553f/karma.conf.js#L42) added support for this type of import statement.

rianfowler: Forgot to mention- I tested this by running the design system. There are a lot of warnings / errors on master that made this a little tricky. will need to clean those up.

PR #54, opened 2018-03-12
add pre-publish script; update readme

rianfowler:


package.json
```diff
@@ -7,10 +7,11 @@
     "type": "git",
     "url": "git+https://github.com/department-of-veterans-affairs/design-system.git"
   },
-  "main": "dist/jeanpants",
+  "main": "dist/jean-pants",
   "scripts": {
     "prestart": "node prebuild.js",
     "export-components": "node ./scripts/build-node-modules.js",
+    "pre-publish": "cp package.json dist/jean-pants/package.json && echo '\\033[36mChange directory to dist/jean-pants\nVerify package.json\nRun command: npm publish'",
```
rianfowler: Quick and dirty pre-publish script. The previous instructions left a gap to this step.
screen shot 2018-03-12 at 12 46 31

package.json
```diff
@@ -7,10 +7,11 @@
     "type": "git",
     "url": "git+https://github.com/department-of-veterans-affairs/design-system.git"
   },
-  "main": "dist/jeanpants",
+  "main": "dist/jean-pants",
   "scripts": {
     "prestart": "node prebuild.js",
     "export-components": "node ./scripts/build-node-modules.js",
+    "pre-publish": "cp package.json dist/jean-pants/package.json && echo '\\033[36mChange directory to dist/jean-pants\nVerify package.json\nRun command: npm publish --access public'",
```
rianfowler: Quick and dirty pre-publish script. The previous instructions left a gap to this step.
screen shot 2018-03-12 at 12 49 43

jbalboni: Is switching directories the only way to get this to work the way we want?

rianfowler:
We could copy the files to `./jean-pants` and build an `.npmignore` for all of the root level files but that folder. Would probably tweak the build to rename `./dist` if we did that.

rianfowler:
I think none of the examples I read / discussed were for npm modules with an organization. If we didn't have `@dva`, the files would have to be copied into the root directory of the project.

PR #55, opened 2018-03-12
Collapsible panel

phiden:


package.json
```diff
@@ -25,7 +24,10 @@
     "foundation-sites": "5",
     "lodash": "^4.5.1",
     "prop-types": "^15.6.0",
-    "react-transition-group": "^1.x",
+    "react": "^15.6.2",
+    "react-dom": "^15.6.2",
+    "react-scroll": "^1.7.7",
```
rianfowler: We should keep these in `devDependencies` unless they break the `Fractal` build. `react` is listed as a `peerDependency` so it's probably not needed in `dependencies` (but it is needed in `devDependencies`).

phiden: Updated!

PR #60, opened 2018-03-14
added AdditionalInfo component to Fract

lbai001:
additionalinfo component added


src/components/AdditionalInfo/AdditionalInfo.unit.spec.jsx
```diff
@@ -0,0 +1,22 @@
...
+  // for the sake of more tests could be written
+  // declared beforeEach and assigned to wrapper
+  beforeEach(() => {
+    wrapper = mount(<AdditionalInfo triggerText="test"/>);
+  });
+
+  it('should render', () => {
+    expect(wrapper.text()).to.contain('test');
+  });
+  it('should pass aXe check', () => {
```
jbalboni: Can we also write tests that run axe on the open state?

src/components/AdditionalInfo/AdditionalInfo.jsx
```diff
@@ -0,0 +1,50 @@
...
+      <ExpandingGroup open expandedContentId={this.expandedContentId}>
+        {trigger}
+        <div className="additional-info-content">{children}</div>
+      </ExpandingGroup>
+    );
+  }
+}
+
+AdditionalInfo.propTypes = {
+  triggerText: PropTypes.string
```
jbalboni: Can you add some documentation to this prop type? You can look at the other components in here for examples. Also, we probably want to make this prop required (probably should have been done in vets-website).

src/components/AdditionalInfo/AdditionalInfo.unit.spec.jsx
lbai001: okay

src/components/AdditionalInfo/AdditionalInfo.unit.spec.jsx
```diff
@@ -0,0 +1,35 @@
...
+    return axeCheck(<AdditionalInfo triggerText="test"/>);
+  });
+  it('renders both children when open is true', () => {
+    const first = wrapper.find('ExpandingGroup').props();
+    expect(first.open).to.eql(true);
```
jbalboni: You could use `.to.be.true` here.

PR #67, opened 2018-03-21
Migrate errorablefileinput

phiden:


src/components/form/controls/ErrorableFileInput/ErrorableFileInput.jsx
```diff
@@ -0,0 +1,121 @@
...
+  errorMessage: PropTypes.string,
+
+  /* string */
+  buttonText: 'Add Files',
+
+  /* file types */
+  mimeTypes: '',
+
+  /* accepts multiple files */
+  multiple: false
```
rianfowler: Remove `buttonText`, `mimeTypes`, `multiple` from the `propTypes` object.

phiden: so if we do that, it breaks the template renderer -- see the conversation in the design system channel -- but would it work to remove the default props instead?

src/components/form/controls/ErrorableFileInput/ErrorableFileInput.unit.spec.jsx
```diff
@@ -0,0 +1,52 @@
...
+    };
+
+    tree.getMountedInstance().handleChange(event);
+
+    expect(onChange.called).to.be.true;
+    expect(event.target.value).to.be.null;
+  });
+
+  it('passes aXe check', () => {
+    const check = axeCheck(<ErrorableFileInput buttonText="my label" onChange={onChange}/>);
```
rianfowler: Probably need to remove the `onChange` prop here.

src/components/form/controls/ErrorableFileInput/ErrorableFileInput.jsx
rianfowler: Oh I see now. 

We can remove `buttonText: 'Add Files,` line- `buttonText` has a `propType` defined for it on line 84. 

In the `propTypes` object, update these two properties: 
`mimeTypes: '',` => `mimeTypes: PropTypes.string,`
`multiple: false` => `multiple: PropTypes.bool`. 

The `defaultProps` object should work correctly the way it is.

phiden: done, thanks for the direction!

phiden:
@rianfowler any idea why Jenkins doesn't like the path to tooltip? It works fine locally.

PR #68, opened 2018-03-22
Migrate errorable date

phiden:


jbalboni:
Lodash is a large dependency for us, even with its babel plugin, so we had started creating some of the utils we use ourselves in order to avoid having the schema form code depend on lodash in the future. If they're not being used in a component, we shouldn't move them over

This is a lot of files, we might want to think about simplifying our date component and seeing if we can reduce the code that's being moved over here.

phiden:
I brought over I think the entire `utils` directory thinking we'd need it later; will write a ticket to remove what's not in active use.

PR #76, opened 2018-03-26
Color palette

phiden:


src/components/colors/colors.njk
```diff
@@ -1,18 +1,20 @@
-{% for family in family %}
+<!-- do you render {{ family }} -->
```
jbalboni: Should this still be here?

sass/style.scss
```diff
@@ -89,3 +89,7 @@
 // ----- VA SHAME ---- //
 
 @import "shame";
+
+// ----- Fractal-specific ---- //
+
+@import "fractal"
```
jbalboni: I don't have a good suggestion for this right now, but we probably will not want to import fractal specific styles in here. I think we'll eventually want to share styles and wouldn't want to pull in fractal ones unnecessarily.

This is fine for now, though, since we have some more work to get the styles to that point anyway.

phiden: Agree -- Fractal has a way to add a fractal-specific stylesheet (that is, affects only the web UI) but I haven't figured out where to put that so that the system sees it. 

I think this may be a different thing, though -- there will be components that are part of Jean Pants that aren't necessarily a thing we have to export, like the color palette, that require CSS to make them play nice. Maybe it's a matter of creating a second set of stylesheets that we somehow exclude from the export, if that's even possible.

PR #77, opened 2018-03-27
adds AcceptTermsPrompt component

lbai001:
adds AcceptTermsPrompt component to Fract
adds AcceptTermsPrompt scss
screen shot 2018-03-26 at 11 35 50 pm


src/components/AcceptTermsPrompt/AcceptTermsPrompt.jsx
```diff
@@ -0,0 +1,146 @@
...
+  componentDidMount() {
+    window.dataLayer.push({ event: 'terms-shown' });
+    window.scrollTo(0, 0);
+  }
+
+  onCancel(e) {
+    e.preventDefault();
+    if (document.referrer !== '' && !this.props.isInModal) {
+      // commented out for design system
+      // browserHistory.goBack();
```
jbalboni: Instead of commenting these out, can we have this component call an `onCancel` prop? We can push the logic here out into the calling components, which shouldn't be too painful.

lbai001: sure thing

sass/style.scss
```diff
@@ -85,7 +85,6 @@
 @import "modules/m-education-wizard";
 @import "modules/m-applications";
 @import "modules/m-additional-info";
+@import "modules/m-accept-terms";
 
-// ----- VA SHAME ---- //
-
-@import "shame";
+// ----- VA SHAME ---- // @import "shame";
```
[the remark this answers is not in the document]
lbai001: ok

PR #80, opened 2018-03-28
updated footer to match vets website

lbai001:
with some link content updates

"Apply for Printed Veteran ID Card" ->"Apply for a printed Veteran ID Card."
"Print Out Your VA Welcome Kit" -> "Print out your VA Welcome Kit."
"Education benefits by school" -> "Check education benefits by school" (it's the only one in that column that doesn't begin with a CTA)


src/components/footer/footer.njk
```diff
@@ -104,6 +137,16 @@ visit <a href="https://www.veteranscrisisline.net/">VeteransCrisisLine.net</a> f
     });
   }
 </script>
+{% if widgets %}
```
jbalboni: We could probably leave this block out. It won't ever include anything here.

PR #81, opened 2018-03-28
Fix build script

jbalboni:
We need to build this in production mode so that the urls are correct on the github.io site.


jbalboni:
@rianfowler It's permanent. The io site is where the published version of the fractal site lives, so you can view it without running the site locally: https://department-of-veterans-affairs.github.io/design-system

PR #82, opened 2018-03-29
Templating

phiden:


createWebpackBundle.js
```diff
@@ -82,7 +86,27 @@ function createWebpackBundle(logger, fractalComponents, watch = true) {
         },
         {
           test: /\.scss$/,
-          use: ExtractTextPlugin.extract({
+          exclude: /\.fractal.scss/,
+          use: componentCSS.extract({
```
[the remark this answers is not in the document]
phiden: It creates a separate CSS file that's specific to the Fractal UI and doesn't affect any of the components.

jbalboni:
@phiden I think I sent you down the wrong path on the separate fractal styles thing, because I didn’t remember how I had set up the sass. For now, I think all we really need to do is to keep that fractal.scss file from being imported in style.scss. But you could import it from site.scss, which is the file I created to import everything for the purposes of the fractal site.

We might still in the future want to do something like the Webpack config changes you have, if we need to provide jean-pants users with css instead of just the sass files. But I don't think that's required right now.

phiden:
@jbalboni possibly you did? But also -- one of the reasons there _is_ fractal-specific CSS is because of the changes I'm making to the system's UI. (though I will move the color-specific styles into site.scss -- I see what you mean now)

That means overriding Mandelbrot's default CSS with a file of our own, and if that file is the `styles.css` that's currently being output, it makes a huge visual mess -- I'll have to then override all the vets-website styles that I don't want applied to the Fractal UI. On the other hand, if I split out the CSS from jump, I can avoid all that.

Is there a reason I'm missing why webpack set up this way is problematic?

jbalboni:
@phiden Ah, ok. I didn't see the other reason for building them separately. That makes sense to me, then. There was nothing specifically wrong with the Webpack changes, they just weren't necessary if our only goal was to keep the fractal sass separate from the component sass.

createWebpackBundle.js
```diff
@@ -82,7 +85,27 @@ function createWebpackBundle(logger, fractalComponents, watch = true) {
...
+            fallback: 'style-loader',
+            use: [
+              { loader: 'css-loader' },
+              { loader: 'resolve-url-loader' },
```
jbalboni: I think you can replace resolve-url-loader and the next loader object with just `{ loader: 'sass-loader' }`. We'll do the same for the config for the other file soon, when we bring the styles back up to date with vets-website.

createWebpackBundle.js
```diff
@@ -27,6 +27,9 @@ function createWebpackBundle(logger, fractalComponents, watch = true) {
 
   fs.writeFileSync('./fractalEntry.js', output);
 
+  const componentCSS = new ExtractTextPlugin('[name].css');
+  const fractalCSS = new ExtractTextPlugin('[name]-fractal.css');
```
jbalboni: I don't think you have to have `[name]` in here, if you want to fix the redundant `fractal-fractal.css` file name.

sass/_shame.scss
```diff
@@ -277,8 +277,190 @@ form {
...
   width:250px;
   height:auto;
   border:1px solid hotpink;
+
+}
+
+// additional variables
+$color-gold-darker: #988530;
+
+.palette-container {
```
jbalboni: I think you mentioned it in a comment, but we should move these out into the fractal only sass.

src/components/form/templates/address-form.njk
```diff
@@ -1,4 +1,4 @@
-
+<div class='usa-grid usa-width-one-whole'>
```
jbalboni: Nitpicky: we normally use double quotes in html.

createWebpackBundle.js
phiden: When I tried this, webpack failed, so I'm going to leave it as-is for the moment.

PR #92, opened 2018-04-04
migrate updates

rianfowler:
Migrated some small updates from the `vets-website`. Will make a separate PR for the new version of the npm module. 

This may break some of the display behavior of the components in the design system. I'm confused how some of these components work right now fex the `ErrorableTextInput` is set up as a controlled component but it retains what's typed in even though no additional `render`s are called. 

The `ErrorableSelect` is also set up as a controlled component but does not render the selected value on interaction.

### DIFF
--- a/src/components/Tooltip/Tooltip.jsx
+++ b/src/components/Tooltip/Tooltip.jsx
@@ -16,24 +16,11 @@ class ToolTip extends React.Component {
     super(props);
     this.handleOpen = this.handleOpen.bind(this);
     this.handleClose = this.handleClose.bind(this);
-    this.state = {
-      closed: !this.props.open
-    };
+    this.state = {};
   }
 
   componentWillMount() {
     this.setState({ closed: true });
-  }
-
-  onBlur(e) {
-    if (!this.state.closed) {
-      const { currentTarget } = e;
-      setTimeout(() => {
-        if (!currentTarget.contains(document.activeElement)) {
-          this.setState({ closed: true });
-        }
-      });
-    }
   }
 
   handleOpen() {
@@ -46,7 +33,7 @@ class ToolTip extends React.Component {
 
   render() {
     return (
-      <div onBlur={this.onBlur}>
+      <div className="hca-tooltip">
         <button tabIndex={this.props.tabIndex} onFocus={this.handleOpen} onClick={this.handleOpen}>More Info</button>
         <div aria-hidden={this.state.closed}>
           {this.props.toolTipText}
@@ -63,13 +50,7 @@ ToolTip.propTypes = {
    * tabIndex for the tooltip toggle
    */
   /* eslint-disable consistent-return */
-  tabIndex(props, propName, componentName) {
-    if (!/^\d+$/.test(props[propName])) {
-      return new Error(`Invalid prop \`${propName}\` supplied to` +
-        ` \`${componentName}\`. Validation failed.`);
-    }
-  /* eslint-enable consistent-return */
-  },
+  tabIndex: PropTypes.number,
   /**
    * Text of the tooltip
    */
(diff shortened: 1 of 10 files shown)


src/components/form/controls/ErrorableCheckbox/ErrorableCheckbox.njk
```diff
@@ -5,7 +5,7 @@ export default function ErrorableCheckboxExample(props) {
   return (
     <ErrorableCheckbox
       label='Errorable Checkbox'
-      onValueChange={(e) => {e.target.checked}}
+      onValueChange={(value) => value}
```
rianfowler: This was throwing an error in the console on the fractal site.

src/components/form/controls/ErrorableRadioButtons/ErrorableRadioButtons.jsx
```diff
@@ -198,6 +201,14 @@ ErrorableRadioButtons.propTypes = {
...
    */
   tabIndex: PropTypes.number,
+  /**
+   * Mouse Down handler
+   */
+  onMouseDown: PropTypes.func,
+  /**
+   * Key Down handler
+   */
+  onKeyDown: PropTypes.func,
```
rianfowler: Two new props added

src/components/Tooltip/Tooltip.jsx
```diff
@@ -16,26 +16,13 @@ class ToolTip extends React.Component {
...
-      closed: !this.props.open
-    };
+    this.state = {};
   }
 
   componentWillMount() {
     this.setState({ closed: true });
   }
 
-  onBlur(e) {
```
rianfowler: This may break the display in the design system. I didn't see anything that looked like a tooltip when I went to test it.

src/components/form/controls/ErrorableSelect/ErrorableSelect.config.yml
```diff
@@ -4,6 +4,7 @@ status: wip
 context:
   componentSourcePath: './ErrorableSelect.jsx'
   errorMessage: "This is the error message"
+  label: 'select label'
```
rianfowler: This was throwing an error on the console of the fractal site

src/components/form/controls/ErrorableSelect/ErrorableSelect.jsx
```diff
@@ -123,6 +124,10 @@ ErrorableSelect.propTypes = {
   // Select field label.
   label: PropTypes.oneOfType([PropTypes.string, PropTypes.element]).isRequired,
 
+  /*
+   * KeyDown handler
+   */
+  onKeyDown: PropTypes.func,
```
rianfowler: new keydown prop for the `ErrorableSelect`

src/components/Tooltip/Tooltip.jsx
jbalboni: I think we need to evaluate where this tooltip is used. I don't think it's ever actually used through the errorable components on the site.

jbalboni:
@rianfowler ErrorableTextInput is working like that because we aren't passing a value, I think. I believe React sees undefined as the value and treats it as uncontrolled. ErrorableSelect does set a value, but there's no state handling in the njk file so it doesn't update. We should probably be consistent about that and do some basic state handling in those njk files.

rianfowler:
@jbalboni Sounds good- I remember doing this before.

src/components/modal/Modal.jsx
```diff
@@ -116,6 +145,29 @@ Modal.propTypes = {
...
+  }),
+  /**
+   * secondary button text and action
+   */
+  secondaryButton: PropTypes.shape({
+    text: PropTypes.string.isRequired,
+    action: PropTypes.func.isRequired,
+  }),
+  /*
+   * Style of modeal alert
```
jbalboni: Typo

PR #97, opened 2018-04-06
9771 publishing improvements

rianfowler:
- added readme for npm module 
- updated the readme / publishing script to include pre-publish step 
- patched version (to publish the readme)

### DIFF
--- a/module-readme.md
+++ b/module-readme.md
@@ -1,0 +1,41 @@
+Version 0.2.0
+
+This module contains reusable react components from [vets-website](https://github.com/department-of-veterans-affairs/vets-website) housed in its design system [repo](https://github.com/department-of-veterans-affairs/design-system).
+
+## Quick start
+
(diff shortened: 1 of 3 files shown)


module-readme.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Quick start
+
+`npm install @department-of-veterans-affairs/jean-pants`
+
+`import @department-of-veterans-affairs/jean-pants/COMPONENTNAME`
+
+See [design system](https://department-of-veterans-affairs.github.io/design-system/) and its [repo](https://github.com/department-of-veterans-affairs/design-system) for more info on components.
+
+## Included components:
```
jbalboni: I'm a little worried about this list getting stale, but I suppose we aren't changing these that much.

rianfowler: We can generate this list with the `pre-publish` script if we start making a bunch of updates or adding components.

PR #100, opened 2018-04-11
Export styles and reorganize

jbalboni:
Sorry, this turned into a giant PR. Several things are going on here:

- I've updated our styles with the latest from vets-website, removed some that shouldn't be in here, and changed the common scss file to be core.scss and not contain module sass. This allows us to be more flexible in how we re-use these styles from vets-website.
- I'm now exporting common styles in the jean-pants module
- I moved our sass code underneath src
- I changed our static assets directory to be /public and changed the build process to output to /public/dist. This lets us check in some static assets like images and have fractal handle copying them over, but still ignore generated code in dist
- I removed a bunch of unused code in src
- I updated a bunch of images in docs to use assetPath, which is required for those images to show up on the github io site.


src/js/export-components.js
```diff
@@ -1,19 +0,0 @@
-const fs = require('fs');
```
rianfowler: Guess I forgot to delete this.

rianfowler:
Does this mean we aren't building any of the fractal stuff into `/dist` anymore?

jbalboni:
@rianfowler Yes. Which reminds me that I probably have some npm module stuff to fix.

rianfowler:
@jbalboni Yes I was just looking that.

jbalboni:
@rianfowler I had one thing to fix, but it doesn't look like I have to change the directory for the npm module process. I left dist in the ignore files and it seems like it could really be any temporary directory, regardless of what fractal uses for its static assets.

If you think I should change it to public/dist I can, but doesn't look like it's required.

PR #103, opened 2018-04-17
[10019-breadcrumb-styles] Moving from the Vets.gov website into design-system

1Copenut:
First commit, adding breadcrumb styles and markup from Vets.gov website. These include all of the styles and adjustments needed to duplicate a breadcrumb list, show/hide. This is the CSS only, the JS should be project-specific.

This issue is a dependency to https://github.com/department-of-veterans-affairs/vets.gov-team/issues/6083. I will not look to merge that one until this is merged and available.

### DIFF
--- a/src/components/navigation/Breadcrumbs/Breadcrumbs.jsx
+++ b/src/components/navigation/Breadcrumbs/Breadcrumbs.jsx
@@ -3,24 +3,35 @@ import React from 'react';
 
 class Breadcrumbs extends React.Component {
   render() {
-    const crumbs = this.props.crumbs;
+    const { crumbs, id } = this.props;
     return (
-      <nav className="va-nav-breadcrumbs">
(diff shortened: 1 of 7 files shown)


src/components/navigation/Breadcrumbs/Breadcrumbs.jsx
```diff
@@ -3,24 +3,35 @@ import React from 'react';
...
   }
 }
 
 Breadcrumbs.propTypes = {
-
   // array should contain objects that contain each breadcrumb's
   // key, href, and plain-text label
-  crumbs: PropTypes.arrayOf(PropTypes.object).isRequired
+  crumbs: PropTypes.arrayOf(PropTypes.object).isRequired,
+  id: PropTypes.string
```
jbalboni: Can you add some explanation to this id prop type? Probably something similar to what you wrote in the readme.

PR #107, opened 2018-04-23
hooked up pagination css

phiden:


jbalboni:
@phiden I fixed this in another PR I just merged, though I didn't make the same shame file change. If that's not necessary, I think we could close this.

phiden:
Copy that -- thanks for the fix! Closing.

PR #108, opened 2018-04-23
Update collapsible panel scrolling

rianfowler:
tested this with `yalc` on `vets-website`


jbalboni:
@rianfowler Seems like we have to add some code to set up the global object in tests?

rianfowler:
@jbalboni Didn't see an obvious place to add global values so I played it safe.

PR #110, opened 2018-04-25
added modal variations explicitly; tried to do it through config but …

phiden:
…it broke _everything._


jbalboni:
@phiden Do you remember what was breaking when you did it through the config? Duplicating the Modal components is really confusing, so I'd like to see if we can fix whatever the issue is with using variants.

phiden:
I will blame it on cold meds that I do not, but let's leave this open and I will figure it out @jbalboni -- you're right; it's more confusing than I anticipated

jbalboni:
@phiden Sounds good. Let me know if it looks like something in the React integration and I can help debug it. It's very possible I didn't account for variants correctly there.

phiden:
@jbalboni Imma blame the cold again -- it's working now; I must have structured the files wrong.

PR #112, opened 2018-04-26
added ErrorableCheckboxes (plural) to jp

phiden:


jbalboni:
Do we need this component? It was created a long time ago I think for the 1990 form and we don't use it for anything anymore (as far as I know).

cehsu:
This is currently used the the disability increase wizard (which is still in a [PR](https://github.com/department-of-veterans-affairs/vets-website/pull/7500)) only. Should we hold off on migrating this unless we see other opportunities to make use of it across other applications?

jbalboni:
Oh, gotcha. I think putting it in here is fine, then. But it looks like it's getting reviewed partially in that other PR, so we may want to hold off until the review is done (or we can just decide to review it thoroughly here instead).

cehsu:
Good idea, we can wait on this until the other is thoroughly reviewed. Thanks for getting this ready @phiden !

cehsu:
@phiden, the [PR](https://github.com/department-of-veterans-affairs/vets-website/pull/7500) that introduces this component has been merged, so the review process on this can be resumed.

src/components/form/controls/ErrorableCheckboxes/ErrorableCheckboxes.jsx
```diff
@@ -0,0 +1,188 @@
...
+        </legend>
+        {errorSpan}
+        {optionElements}
+      </fieldset>
+    );
+  }
+}
+
+/**
+ * `toolTipText` - String with help text for user.
```
jbalboni: Should we move these down to the prop types?

src/components/form/controls/ErrorableCheckboxes/ErrorableCheckboxes.jsx
```diff
@@ -0,0 +1,188 @@
...
+ * `tabIndex` - Number for keyboard tab order.
+ * `options` -
+ * `required` - is this field required.
+ * `values` - object. Values of the checkbox field.
+ * `onValueChange` - a function with this prototype: (newValue)
+ */
+
+ErrorableCheckboxes.propTypes = {
+  // String for any additional fieldset classes.
```
jbalboni: These need to be in `/* */` form to show up properly (as I just learned recently).

src/components/form/controls/ErrorableCheckboxes/ErrorableCheckboxes.config.yml
```diff
@@ -0,0 +1,19 @@
+label: ErrorableCheckboxes
+status: wip
```
jbalboni: I don't think we need this status, though maybe @phiden has an opinion?

jbalboni:
Oh, oops, I got confused about who was the PR author and thought @cehsu was giving it the ok for review! I can definitely take another look once it's updated.

phiden:
@cehsu could you take a look at this please & let me know if it's good to merge?

src/components/form/controls/ErrorableCheckboxGroup/ErrorableCheckboxGroup.config.yml
```diff
@@ -0,0 +1,19 @@
+label: ErrorableCheckboxGroup
+status: wip
```
cehsu: Is this still the case?

src/components/form/controls/ErrorableCheckboxGroup/ErrorableCheckboxGroup.njk
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import ErrorableCheckboxGroup from './ErrorableCheckboxGroup';
+
+export default function ErrorableCheckboxGroupExample(props) {
+  return (
+    <ErrorableCheckboxGroup
+      label='Errorable Checkbox Group'
```
cehsu: It looks like there are some "required" props are not provided here (this seems likely), should they be added here or should the proptypes be updated?

phiden: it looks like they're being provided by the config file?

src/components/form/controls/ErrorableCheckboxGroup/ErrorableCheckboxGroup.config.yml
phiden: nope; will update

PR #117, opened 2018-05-11
rename module references to formation

rianfowler:
Will publish formation once this is merged and work on migrating vets-website to it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
-  "name": "@department-of-veterans-affairs/jean-pants",
-  "version": "0.6.1",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "@department-of-veterans-affairs/jean-pants",
-  "version": "0.6.1",
+  "name": "@department-of-veterans-affairs/formation",
+  "version": "0.7.0",
```
rianfowler: Opted not to restart the version count. Dunno if this is the preferred way.

PR #118, opened 2018-05-14
hotfix/7724: fix for radiobox issue with none component radioboxes

accbjt:
This is a fix for none component radio button not working due to positioning issues.

How to test:

ErrorableRadioButton should still look like this on focus.
screen shot 2018-05-14 at 2 07 23 pm

On vets-website the radio buttons should be selectable and working. Also it should not have the yellow highlight. Use this link to check the radio button form.

http://localhost:3001/health-care/apply/application/household-information/financial-disclosure?skip

This is what it should look like now:
screen shot 2018-05-14 at 2 09 58 pm

### DIFF
--- a/src/components/form/controls/ErrorableRadioButtons/ErrorableRadioButtons.jsx
+++ b/src/components/form/controls/ErrorableRadioButtons/ErrorableRadioButtons.jsx
@@ -98,7 +98,7 @@ class ErrorableRadioButtons extends React.Component {
         <div
           key={optionAdditional ? undefined : optionIndex}
           className="form-radio-buttons">
-          <div className="errable-radio-button">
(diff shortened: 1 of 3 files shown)


src/components/form/controls/ErrorableRadioButtons/ErrorableRadioButtons.jsx
```diff
@@ -98,7 +98,7 @@ class ErrorableRadioButtons extends React.Component {
         <div
           key={optionAdditional ? undefined : optionIndex}
           className="form-radio-buttons">
-          <div>
+          <div className="errable-radio-button">
```
jbalboni: I think this should be `errorable` instead of `errable`, to be consistent.

PR #119, opened 2018-05-16
Update primary color to avoid aXe issue

jbalboni:
Our saved application status widget is generating an aXe error:

screen shot 2018-05-16 at 11 33 36 am

screen shot 2018-05-16 at 1 12 32 pm

This change very slightly tweaks the primary color to get above the WCAG threshold. I can't tell a difference visually. The changed color passes the contrast check: https://snook.ca/technical/colour_contrast/colour.html#fg=0071BB,bg=E1F3F8

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,12 +1,12 @@
 {
   "name": "@department-of-veterans-affairs/formation",
-  "version": "0.7.1",
+  "version": "0.7.2",
   "license": "CC0-1.0",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/department-of-veterans-affairs/design-system.git"
   },
-  "main": "dist/jeanpants",
+  "main": "dist/formation",
   "scripts": {
     "prestart": "node prebuild.js",
     "export-components": "node ./scripts/build-node-modules.js",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,12 +1,12 @@
...
   "name": "@department-of-veterans-affairs/formation",
-  "version": "0.7.1",
+  "version": "0.7.2",
   "license": "CC0-1.0",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/department-of-veterans-affairs/design-system.git"
   },
-  "main": "dist/jeanpants",
+  "main": "dist/formation",
```
jbalboni: Saw this, but I don't think it's affecting anything.

src/sass/base/_b-variables.scss
```diff
@@ -31,6 +31,8 @@ $site-max-width:      100rem; // Worksout to about 1000px. USWDS value is 1040px
 //===================================
 
 $color-base:                 #212121;
+$color-blue:                 #0071BB;
```
jbalboni: Default color is #0071BC

package.json
rianfowler: I think we can remove this. `main` is 
- explicitly whitelisted file that's always added to the module when it's published, and 
- defines the entry point for the module when it's imported. 

Since there's no single entry point to this module and we're requiring users to directly import the components they use, we should probably just remove it.

PR #121, opened 2018-05-18
Fix Modal TypeError

ncksllvn:
If the modal opens without a previous element ever having focus, there is a `TypeError` occurring in `componentWillReceiveProps`. This happens in the case of the DowntimeNotifications, which opens a modal upon page load if downtime is approaching. In that case, clicking the close button will result in an error and won't close the modal.

image

### DIFF
--- a/src/components/modal/Modal.jsx
+++ b/src/components/modal/Modal.jsx
@@ -40,7 +40,9 @@ class Modal extends React.Component {
     } else if (!newProps.visible && this.props.visible) {
       document.removeEventListener('keyup', this.handleDocumentKeyUp, false);
       document.removeEventListener('focus', this.state.focusListener, true);
-      this.state.lastFocus.focus();
+      if (this.state.lastFocus) {
+        this.state.lastFocus.focus(); 
+      }
       document.body.classList.remove('modal-open');
     }
   }
--- a/src/components/modal/Modal.jsx
+++ b/src/components/modal/Modal.jsx
@@ -43,7 +43,7 @@ class Modal extends React.Component {
     } else if (!newProps.visible && this.props.visible) {
       document.removeEventListener('keyup', this.handleDocumentKeyUp, false);
       document.removeEventListener('focus', this.state.focusListener, true);
-      this.state.lastFocus.focus(); 
+      this.state.lastFocus.focus();
       document.body.classList.remove('modal-open');
     }
   }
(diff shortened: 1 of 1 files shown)


src/components/modal/Modal.jsx
```diff
@@ -40,7 +40,9 @@ class Modal extends React.Component {
     } else if (!newProps.visible && this.props.visible) {
       document.removeEventListener('keyup', this.handleDocumentKeyUp, false);
       document.removeEventListener('focus', this.state.focusListener, true);
-      this.state.lastFocus.focus();
+      if (this.state.lastFocus) {
```
jbalboni: After the modal is closed for downtime notifications, what happens if you start tabbing? Does it start from the top of the page? We want to make sure screen reader users aren't stuck in some random part of the page.

ncksllvn: The modal is intended to be shown at application startup, so I think the top of the page makes sense. 

Also, the DowntimeNotification's modal implementation could possibly use a rework, because it's being rendered conditionally using state, rather than using the `visible` prop on the Modal.

ncksllvn:
Here's the [code](https://github.com/department-of-veterans-affairs/vets-website/pull/7757/files#diff-8a63aa7026b7b5f3e0ce111bbf063cc7R31) where I'm seeing this error. I think I'm falling the best practices there so I'm not sure what could be the issue. If it's being used correctly, can we merge this soon?

jbalboni:
@ncksllvn I still have the same question I asked above: with this change, where does focus end up when you clear that modal on the home page? I think you need to do something in an `else` condition if it's not at the top of the page. If that is the case, though, this looks good to me.

ncksllvn:
@jbalboni Oh my bad. Since the modal will render before any interactivity with the application I think it makes sense to focus on the top of the page. I may be misunderstanding your question still though.

jbalboni:
Right, I think we're in agreement. I just don't know how to test this change out to see if that's happening now, so I was wondering if you had. We'd want to change the code in this modal component to specifically set the focus to the top of the page when `lastFocus` is falsey if it's not working with this change.

ncksllvn:
@jbalboni Would defaulting `state.lastFocus` to `document.activeElement` be a good change? That would remove the need for that safety check.

jbalboni:
@ncksllvn Yeah, I think that would work.

ncksllvn:
@jbalboni Removed that safety check and instead added that default value for lastFocus. PTAL when you get the chance.

PR #122, opened 2018-05-22
Minor updates to AlertBox

jbalboni:
Makes some minor updates to AlertBox to make it more configurable. No visual changes.

### DIFF
--- a/src/components/alerts/AlertBox/AlertBox.jsx
+++ b/src/components/alerts/AlertBox/AlertBox.jsx
@@ -39,7 +39,8 @@ class AlertBox extends React.Component {
 
     const alertClass = classNames(
       'usa-alert',
-      `usa-alert-${this.props.status}`
+      `usa-alert-${this.props.status}`,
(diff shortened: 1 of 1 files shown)


src/components/alerts/AlertBox/AlertBox.jsx
```diff
@@ -39,7 +39,8 @@ class AlertBox extends React.Component {
 
     const alertClass = classNames(
       'usa-alert',
-      `usa-alert-${this.props.status}`
+      `usa-alert-${this.props.status}`,
+      this.props.className
```
jbalboni: Adding this so we can more easily adjust things like margins.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation",
-  "version": "0.7.2",
+  "version": "0.8.2",
```
rianfowler: `0.8.0`

PR #123, opened 2018-05-22
Feature/6082

accbjt:
This need to be merged and deployed before the pull request below gets approved.

https://github.com/department-of-veterans-affairs/vets-website/pull/7676


src/sass/base/focus.scss
```diff
@@ -3,3 +3,51 @@
...
+input[type="radio"]:focus {
+  & + label {
+    &::before {
+      @include focus-gold-light-outline;
+      position: relative;
+      z-index: 3;
+    }
+  }
+
+  & ~ div {
```
jbalboni: What's this rule doing? Using a `div` element selector makes a little wary.

jbalboni: Isn't the sibling of the radio input a label?

phiden:
@jbalboni @accbjt what's the status of this ticket?

accbjt:
@phiden I will look at it today. There was a reason why I did that but I need to look into it. I have to check the radio buttons that are not React components and see if that effects it or not. I will check this out today so we can merge the code in.

src/sass/base/focus.scss
accbjt: I removed this. I tried to find an input that is not an ErrorableRadioButton. This was a catch all which would select any radio buttons that where not ErrorableRadioButton. I couldn't find an example in the current codebase, so I am removing this styling. If this is okay we can merge the current styling so it reflects the new on focus stylings.

phiden:
@accbjt what's the status of this ticket?

accbjt:
@phiden I removed the css @jbalboni was commenting on. It doesn't seem like it's relevant with the new codebase. If everything looks good to you, we can merge it into master.

PR #124, opened 2018-06-15
11002-breadcrumb-component

1Copenut:
## Bumped Version
* Bumped version number to 1.0.0 due to the complete restructuring and risk of breakage.
* The previous version used a `props.children` object to build list items and links, but assumed too much about the structure of those children.
* New version allows hard coded links, React Router `<Link />` or arrays of one, the other, or both

## Added
* A fully revamped `<Breadcrumbs />` component, as part of the continuing effort to split my pull request for [Vets-website #6083](https://github.com/department-of-veterans-affairs/vets-website/pull/7579) into smaller, more manageable chunks
* The breadcrumbs component has several new props for adding your own `aria-label`, custom CSS classes, and custom mobile width breakpoint
* Documentation to match

## Removed
* `{...props}` for now, because there's not a clear case for it, and React complains about non-DOM-specific props on rendered HTML. I did factor out the mobile-width to a `data-mobile-width` so future selves can reference it.
* Will add back the destructed props if there's a need during actual code use


src/components/navigation/Breadcrumbs/Breadcrumbs.jsx
```diff
@@ -1,40 +1,143 @@
...
+      id,
+      listId
+    } = this.props;
+    const breadcrumbId = id || uniqueId('va-breadcrumbs-');
+    const breadcrumbListId = listId || uniqueId('va-breadcrumbs-list-');
+    const mobileShow = this.state.mobileShow;
+    const shownList = mobileShow
+      ? (
+        <ul
+          className="row va-nav-breadcrumbs-list columns"
```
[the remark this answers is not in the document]
1Copenut: @jbalboni Yes, it's offering some padding that looks like it's baked into the USWDS media queries, so removing it causes layout shift.

src/helpers/debounce.js
```diff
@@ -0,0 +1,32 @@
...
+
+    // Execute after timeout perioud
+    timeout = setTimeout(later, wait);
+
+    // If callNow evaluates true, call func, then wait for timeout
+    if (callNow) {
+      func.apply(this, args);
+    }
+  };
+}
```
ncksllvn: Is there an advantage to using this over the Lodash [debounce](https://lodash.com/docs/4.17.10#debounce) function? Or is is because we're in the Design System?

src/components/navigation/Breadcrumbs/Breadcrumbs.jsx
```diff
@@ -1,40 +1,152 @@
...
+  customClasses: PropTypes.string,
+  /* Adds a custom id attribute to the <nav> element */
+  id: PropTypes.string,
+  /* Adds a custom id attribute to the <ul> element */
+  listId: PropTypes.string,
+  /* Causes the mobile back by one link to be shown while TRUE */
+  mobileFirstProp: PropTypes.bool,
+  /* Changes the viewport width that will toggle full and mobile breadcrumb UI. Defauls to 481 device pixels. */
+  mobileWidth: PropTypes.number.isRequired,
 };
```
ncksllvn: I think we're aiming to provide prop descriptions as JSDoc comments, such as with [here](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/platform/monitoring/DowntimeNotification/containers/DowntimeNotification.jsx#L24).

src/components/navigation/Breadcrumbs/Breadcrumbs.jsx
```diff
@@ -1,40 +1,152 @@
...
       </nav>
     );
   }
 }
 
+Breadcrumbs.defaultProps = {
+  ariaLabel: 'Breadcrumb',
+  mobileWidth: 481,
+};
+
```
ncksllvn: Nice!

src/helpers/debounce.js
1Copenut: @ncksllvn Not really, other than me missing Lodash was already a dependency. Can switch it and save us a few bytes on the bundle file. Good catch!

1Copenut:
@ncksllvn Made the changes you suggested. I also moved the IDs into class declaration, because they were being updated every time `render()` fired, which seemed excessive. Will merge shortly if no objections.

PR #125, opened 2018-06-20
Remove forms-related styles

annekainicUSDS:
In this PR I am moving forms-related styles that are only used within us-forms-system to us-forms-system. They are moved to us-forms-system in [this PR](https://github.com/usds/us-forms-system/pull/91). These styles are not referred to elsewhere within VA design system.


src/sass/modules/_m-form-elements.scss
```diff
@@ -22,22 +22,25 @@
       }
     }
   }
+
+  > input[type="radio"] + label::before {
```
annekainicUSDS: These styles were moved up because they were included later in this file under a separate reference to `.form-radio-buttons`. I consolidated the references to just 1 and included these styles here.

src/sass/modules/_m-form-elements.scss
```diff
@@ -199,16 +166,81 @@ button.form-button-disabled {
     content: "-";
 }
 
-.form-errorable-group {
-  > .usa-input-error {
-    label {
-      margin-top: 3rem;
+// Overriding the accordion button styles that are too broad
```
annekainicUSDS: These styles were copied from the file `_m-form-process`. They were the only styles that were still needed in this repo from that file, so it seemed that consolidating all the forms related stuff into one file would work so I could delete the other one.

jbalboni:
@annekainicUSDS Is the plan for us to import `us-forms-system` styles in `vets-website` to replace these?

annekainicUSDS:
@jbalboni yes I think we'll need to, so perhaps we merge this PR I should do a bit of refactoring on vets-website as well

jbalboni:
With these changes, how do the form components in here look on the Fractal site? Are we breaking their styling? I wonder if we should wait to remove styles in here until the styling in us-forms-system is further along.

annekainicUSDS:
Totally fair @jbalboni. I'll close this PR for now until I'm able to sort things out a bit more between us-forms-system, vets-website, and here.

PR #126, opened 2018-06-28
Feature/10551

accbjt:
This is for the new MegaMenu component for the merger work on va.gov.


src/components/navigation/MegaMenu/MegaMenu.njk
```diff
@@ -0,0 +1,797 @@
+import React from 'react';
+import MegaMenu from './MegaMenu';
+
+const data = [
```
jbalboni: I'm curious what @phiden things, but I don't think we necessarily have to have the full menu (or almost the full menu) in here. Presumably this is also defined where these component is used in vets-website and this info will end up getting stale after a while.

src/components/navigation/MegaMenu/MegaMenu.unit.spec.jsx
```diff
@@ -0,0 +1,745 @@
+import React from 'react';
+import { expect } from 'chai';
+import sinon from 'sinon';
+import { mount } from 'enzyme';
+
+import MegaMenu from './MegaMenu.jsx';
+
+const data = [
```
jbalboni: Similar to above, you could probably use shorter dummy data for the menu and make this test a little easier to read.

src/components/navigation/MegaMenu/MegaMenu.unit.spec.jsx
```diff
@@ -0,0 +1,745 @@
...
+
+  it('should show Disability section if state.currentSection is = Disability', () => {
+    megaMenu.setProps({
+      ...megaMenu.props(),
+      currentDropdown: 'Health and Benefits',
+      currentSection: 'Disability',
+    });
+
+    expect(megaMenu.find('.panel-title').at(0).text()).to.equal('Get Disability Benefits');
+  });
```
jbalboni: We also try to write tests that run our axeCheck for all the visual states of the component (menus closed, menus open, etc). You can search for `axeCheck` to find examples.

src/sass/modules/_m-megamenu.scss
```diff
@@ -0,0 +1,115 @@
+@import "../shared-variables";
```
jbalboni: Was not having this causing you trouble somewhere? It shouldn't be necessary if this file is used in the same way as the other module sass files.

src/sass/modules/_m-megamenu.scss
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    .vetnav-level2[aria-expanded="true"] {
+      box-shadow: -3px 0 $color-gold-light;
+    }
+
+    .vetnav-panel {
+      left: 0;
+
+      a {
```
jbalboni: I think you should consider writing classes for these instead of directly styling the `a` and `li` elements in this file. Either that or using `>` to restrict them more, if they're a direct descendent (though it doesn't look like they would be). I have found that nesting element selectors leads to problems later with styles that apply to more elements than they should (our accordion styles are a good example of this problem).

src/components/navigation/MegaMenu/MegaMenu.jsx
```diff
@@ -0,0 +1,92 @@
...
+
+export default class MegaMenu extends React.Component {
+  handleOnClick(title) {
+    if (this.props.currentDropdown === title) {
+      this.props.handleOnClick('');
+    } else {
+      this.props.handleOnClick(title);
+    }
+  }
+
```
ncksllvn: I'm not sure how I feel about the name `handleOnClick` because there are a lot of elements within this component so it's kind of unclear what it's referring to until you trace it back. Is there possibly a more descriptive name? Maybe something like `toggleItemMenu` or `handleMenuButtonClick`?

src/components/navigation/MegaMenu/SubMenu.jsx
```diff
@@ -0,0 +1,85 @@
...
+              <li className="panel-title">{mainColumn.title}</li>
+              { mainColumn.links.map((link, i) => (
+                <li key={`${link.href}-${i}`}><a href={link.href}>{link.text}</a></li>
+              ))}
+            </ul>
+          }
+          <ul
+            className={`vetnav-panel vetnav-panel--submenu panel-1 ${mainColumn && 'panel-white'}`}
+            id="vetnav-disability"
+            role="menu"
```
ncksllvn: Should there be an `id` here? Not sure if that was intentional or not.

src/components/navigation/MegaMenu/SubMenu.jsx
```diff
@@ -0,0 +1,85 @@
...
+            aria-label="Disability">
+            <img src={columnThree.img.src} alt={columnThree.img.alt}></img>
+            <a href={columnThree.link.href}>{columnThree.link.text}</a>
+            <p>{columnThree.description}</p>
+          </div>
+        </div>
+      );
+    }
```
ncksllvn: Again possibly just a personal preference, but there are a lot of conditions here and I think it might be nice to see what logic can be abstracted into a child component or helper function. Possibly `<ColumnTwoSubMenu/>`, `<ColumnThreeSubMenu/>` or something like that.

src/components/navigation/MegaMenu/MegaMenu.njk
accbjt: @jbalboni this data will be eventually grabbed from an api request. Not sure if it will be everything or just partial. We want users to be able to change the label and link but not be able to touch the codebase. This will be for content specialist on the va.gov side. We eventually want va.gov people to edit the TeamSite CMS on va.gov and we make a request for the page it will render, which will be json, then we will update our MegaMenu. We are still working out the details but this give us something close to a proof of concept. I can shorten this to have placeholder data for an example.

src/components/navigation/MegaMenu/MegaMenu.jsx
```diff
@@ -0,0 +1,92 @@
...
+            <ul id="vetnav-menu" role="menubar">
+              <li><a href="/" className="vetnav-level1" role="menuitem">Home</a></li>
+              {
+                data.map((item, i) => {
+                  return (
+                    <li key={`${item.title.toLowerCase().replace(/ /g, '-')}-${i}`}>
+                      {
+                        item.menuSections ? <button
+                          aria-expanded={currentDropdown === item.title}
+                          aria-controls="vetnav-explore"
```
cvalarida: :thinking: I think this will have to be based on the menu title.

src/components/navigation/MegaMenu/MegaMenu.jsx
```diff
@@ -0,0 +1,92 @@
...
+                          aria-haspopup="true"
+                          className="vetnav-level1"
+                          onClick={() => this.handleOnClick(item.title)}>{item.title}</button>
+                          : <a href={item.href} className="vetnav-level1" id="pgdpevffu88i">{item.title}</a>
+                      }
```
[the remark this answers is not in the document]
accbjt: I didn't want any name clashes because it wasn't unique enough. If any of the titles where the same we would only get one item rendered. So by using the index we would get a unique key. It's not the best approach but works for this. I was trying to use as much as possible from the older code, so I must have put this one in. Thanks for catching that.

src/components/navigation/MegaMenu/MegaMenu.jsx
accbjt: That is a great suggestion

src/components/navigation/MegaMenu/MegaMenu.njk
```diff
@@ -0,0 +1,797 @@
+import React from 'react';
```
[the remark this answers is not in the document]
accbjt: This is what we use to render our components in the Formation style guide. This lets us create a really nice guide that we can see in the browser.

src/components/navigation/MegaMenu/MenuSection.jsx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+MenuSection.propTypes = {
+  title: PropTypes.string.isRequired,
+  updateCurrentSection: PropTypes.func.isRequired,
+  links: PropTypes.object.isRequired,
+  defaultSection: PropTypes.string.isRequired,
+};
+
+MenuSection.defaultProps = {
```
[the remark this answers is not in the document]
accbjt: No I am removing it.

src/components/navigation/MegaMenu/MegaMenu.njk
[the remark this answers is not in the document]
jbalboni: Yeah, this is a quirk of Fractal. We use njk for plain html pattern examples, but Fractal doesn't let you use two different template engines by default. So for React component examples we have to use the njk extension to make Fractal happy and treat it as JSX in our custom templating code. It's annoying.

src/components/navigation/MegaMenu/MegaMenu.jsx
1Copenut: `aria-controls` must only appear when the ID it references is present, otherwise it throws a Critical axe-core error. It's not technically wrong, but I've found it better to use conditional logic like the following. Assuming `menu1.attributes` are pseudo-code.

```js
aria-controls={ menu1.visible ? ${menu1.id} : null }
```

This creates more work, but also allows for local overrides and flexibility.

1Copenut:
I will have some further comments for this PR. I want to review it with a couple of screen readers, and scan the markup with axe.

src/components/navigation/MegaMenu/MegaMenu.njk
1Copenut: I think it will be good to have a shorter version for proof of concept, but also this full version for user analysis testing. I have some questions about having the first menu (health care) open by default. As a user, I generally like to tab through things and make a selection for more content. Screen reader users would have to tab through all of the `health care` links to hear "Disability, menu item button". That's an experience we want to test.

src/components/navigation/MegaMenu/MenuSection.jsx
```diff
@@ -0,0 +1,37 @@
...
+class MenuSection extends React.Component {
+  getCurrentSection(props) {
+    return props.currentSection ? props.currentSection : props.defaultSection;
+  }
+
+  render() {
+    const show = this.getCurrentSection(this.props) === this.props.title;
+
+    return (
+      <li>
```
1Copenut: These first level `<li>` elements need a `role="menuitem"` or they'll throw an axe-core error, and announce themselves inconsistently to screen readers. See attached screenshot.

---

screen shot 2018-06-29 at 2 12 53 pm

src/components/navigation/MegaMenu/SubMenu.jsx
```diff
@@ -0,0 +1,85 @@
...
+              className={`vetnav-panel vetnav-panel--submenu panel-2 ${mainColumn && 'panel-white'}`}
+              id="vetnav-disability"
+              role="menu"
+              aria-label="Disability">
+            </ul>
+          }
+          <div
+            className="vetnav-panel vetnav-panel--submenu panel-3"
+            id="vetnav-disability"
+            role="menu"
```
1Copenut: This `<div>` is not a menu, just a container for links. I think removing the `role="menu"` is in order.

src/components/navigation/MegaMenu/SubMenu.jsx
```diff
@@ -0,0 +1,85 @@
...
+    const { mainColumn, columnOne, columnTwo, columnThree, seeAllLink } = this.props.data;
+
+    if (this.props.show) {
+      return (
+        <div>
+          {
+            mainColumn && <ul
+              className="vetnav-panel vetnav-panel--submenu panel-main"
+              id="vetnav-disability"
+              role="menu"
```
1Copenut: This `<ul>` is not a menu, just a container for links. I think removing the `role="menu"` is in order.

src/components/navigation/MegaMenu/SubMenu.jsx
```diff
@@ -0,0 +1,85 @@
...
+              aria-label="Disability">
```
1Copenut: A better approach might be to remove this first `<li>`, make it a heading, and associate the list with it via `aria-labelledby`. You'll probably need to wrap the lists in another div, which adds a layer, but offers a better semantic relationship.

```html
<div>
  <h3 id="vetnav-disability-header">Get Health Care Benefits</h3>
  <ul aria-labelledby="vetnav-disability-header" id="vetnav-disability">
    ...
    <li>
  </ul>
</div>
```
This allows assistive device users to navigate inside the expanded menu via heading, and gives the list a contextual heading that's read out.

src/components/navigation/MegaMenu/MegaMenu.njk
accbjt: Having health care open by default is the way our vets.gov nav menu works now. I was just using what was already there. I actually had the same feeling when I first was implementing it, but I thought it would be safe to just do it like it is today. @1Copenut should we ask someone about this? It shouldn't be hard to change.

phiden: +1 to shortening as an example; the full menu is massive and I suspect will be highly variable for a long time to come

phiden:
👏 for continued efforts to make this so excellent. Well done, everyone.

accbjt:
@ncksllvn @1Copenut @phiden @jbalboni @cvalarida I made requested changes. Can you please review the changes. Let me know if I missed anything or if you see anything that needs changing.

src/components/navigation/MegaMenu/Column.jsx
```diff
@@ -0,0 +1,65 @@
...
+Column.propTypes = {
+  data: PropTypes.shape({
+    title: PropTypes.string,
+    links: PropTypes.arrayOf(
+      PropTypes.shape({
+        text: PropTypes.string.isRequired,
+        href: PropTypes.string.isRequired,
+      })
+    ),
+    img: PropTypes.shape({
```
ncksllvn: This is super useful!

src/components/navigation/MegaMenu/MegaMenu.jsx
```diff
@@ -0,0 +1,116 @@
...
+                          className="vetnav-level1"
+                          onClick={() => this.toggleDropDown(item.title)}>{item.title}</button>
+                          : <a href={item.href} className="vetnav-level1" >{item.title}</a>
+                      }
+                      <div id="vetnav-explore" className="vetnav-panel" role="none" hidden={currentDropdown !== item.title}>
+                        {
+                          item.title === currentDropdown && item.menuSections && <ul aria-label="Explore benefits">
+                            {
+                              item.menuSections.constructor.name === 'Array' ? item.menuSections.map((section, j) => {
+                                return (
```
ncksllvn: Is it possible for `item.title === currentDropdown` but for `item` not to have `menuSections`?

src/components/navigation/MegaMenu/MenuSection.jsx
```diff
@@ -0,0 +1,71 @@
...
+    }),
+    columnTwo: PropTypes.shape({
+      title: PropTypes.string.isRequired,
+      links: PropTypes.arrayOf(
+        PropTypes.shape({
+          text: PropTypes.string.isRequired,
+          href: PropTypes.string.isRequired,
+        })
+      ),
+    }),
```
ncksllvn: Might be worth storing the whole column-shape in a constant and referring to that as the prop-type for each column.

src/components/navigation/MegaMenu/Column.jsx
```diff
@@ -0,0 +1,65 @@
...
+      <h3 id={`vetnav-${strToDash(navTitle)}-header`}>{data.title}</h3>
+      <ul id={`vetnav-${strToDash(navTitle)}`} aria-labelledby={`vetnav-${strToDash(navTitle)}-header`}>
+        { data.links.map((link, i) => (
+          <li className="mm-link-container" key={`${link.href}-${i}`}><a className="mm-links" href={link.href}>{link.text}</a></li>
+        ))}
+
+        {props.children}
+      </ul>
+    </div>
+  );
```
ncksllvn: Much cleaner!

src/components/navigation/MegaMenu/MegaMenu.jsx
ncksllvn: @1Copenut does the ID here meet the criteria for your comment on `aria-controls`?

src/components/navigation/MegaMenu/Column.jsx
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const toDash = (keyName) => {
+  return keyName.replace(/([A-Z])/g, (str) => { return `-${str.toLowerCase()}`;});
+};
+
```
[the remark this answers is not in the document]
accbjt: I think it might. Totally forgot about lodash.

src/components/navigation/MegaMenu/MegaMenu.jsx
accbjt: It should always have a menuSection. I will add a isRequired in my proptypes.

src/components/navigation/MegaMenu/MenuSection.jsx
accbjt: Never thought of doing that. I will try it out.

accbjt:
@ncksllvn We can't do the API call on build unless it can be instantly updated when the API data is updated. We will need this section to be updated by va.gov people and it needs to appear instantly after they have edited the content on their system.

src/components/navigation/MegaMenu/MegaMenu.jsx
```diff
@@ -0,0 +1,116 @@
...
+                                  <MenuSection
+                                    key={`${section}-${j}`}
+                                    title={section.title}
+                                    defaultSection={defaultSection(item.menuSections)}
+                                    currentSection={currentSection}
+                                    updateCurrentSection={() => this.updateCurrentSection(section.title)}
+                                    links={section.links}></MenuSection>
+                                );
+                              }) : <SubMenu data={item.menuSections} navTitle={item.title} show></SubMenu>
```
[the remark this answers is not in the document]
accbjt: I just needed to wrap this with a li tag and it should be good.

ncksllvn:
@accbjt Gotcha, I didn't realize that was a requirement. It might be worth considering doing both - binding the component to generated HTML from the most recent deployment, then hitting the API route and updating if necessary.

ncksllvn:
Once @1Copenut's expectations for accessibility are met I think we're 👍 for integration into the site, although I imagine tweaks to this component will be made during that process.

src/components/navigation/MegaMenu/Column.jsx
```diff
@@ -0,0 +1,65 @@
...
+        <a className="mm-links" href={data.link.href}>{data.link.text}</a>
+        <p>{data.description}</p>
+      </div>
+    );
+  }
+
+  return (
+    <div className={`vetnav-panel vetnav-panel--submenu ${toDash(keyName)} ${panelWhite && 'panel-white'}`}>
+      <h3 id={`vetnav-${strToDash(navTitle)}-header`}>{data.title}</h3>
+      <ul id={`vetnav-${strToDash(navTitle)}`} aria-labelledby={`vetnav-${strToDash(navTitle)}-header`}>
```
1Copenut: We can remove the id from the UL here, so it references the H3 id.

```javascript
<ul aria-labelledby={`vetnav-${strToDash(navTitle)}-header`}>
```

src/components/navigation/MegaMenu/MegaMenu.jsx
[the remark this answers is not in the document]
1Copenut: That looks good to me. Wrapping it as a `<li>` puts it in the unordered list expected format, and is semantically correct.

1Copenut:
@accbjt I have one final adjustment to remove an extra ID, otherwise no additional changes.

https://github.com/department-of-veterans-affairs/design-system/pull/126/files#r203481711

1Copenut:
The mega menu tests out well in the following 508 considerations:

* automated (axe plugin for Chrome): 0 issues
* keyboard navigation: 0 issues
* zoom to 400%: 0 issues, layout adapts to mobile-first gracefully
* 8 unique colorblindness types: 0 issues
* NVDA screenreader + Firefox + Win7: 0 issues
* VoiceOver screenreader + Safari + MacOS 10.13.6 High Sierra latest: 0 issues

I'll evaluate the latest version against IE11 + JAWS in the days ahead and will ticket any findings.

accbjt:
@1Copenut Thanks

PR #130, opened 2018-07-04
Allow AdditionalInfo to handle onClick

bshyong:


src/components/AdditionalInfo/AdditionalInfo.jsx
```diff
@@ -13,6 +13,7 @@ export default class AdditionalInfo extends React.Component {
 
   toggle = () => {
     this.setState({ open: !this.state.open });
+    return this.props.onClick && this.props.onClick();
   };
```
ncksllvn: I know we're adding this here just as quick fix so parent components can respond, but it does sort of seem like the "open" value in state should instead be up a prop passed from the parent.

PR #131, opened 2018-07-05
11398-breadcrumb-css

1Copenut:
## Purpose of this PR

The `<Breadcrumb />` component was making too many assumptions about the shape of its children. By using Javascript to pluck out the child, inline click handlers (analytics, other inline functions) ran the risk of being picked up. This new PR simplifies the Breadcrumb's mobile/full display by making it a CSS decision only.

I will bump the version number MINOR as I consider this to be a non-breaking API change. Users have all of the functionality as before, except the `aria-label` that was previously applied to the mobile-first crumb. In local testing, this seemed to not be a major loss. Will keep an ear out for user feedback and add back as needed later.

## Removed

* A number of unnecessary lifecycle methods, local state
* `mobileWidth` prop - this is now determined by CSS media query only

## Added

* Logic for adding a `va-nav-breadcrumbs--mobile` CSS class for always rendering the mobile view
* CSS for handling the breadcrumb transition as a media query, or CSS class
* CSS uses the `:nth-last-child` selector to hide all breadcrumb links, show just the second-to-last: https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-last-child
* An inline click handler in unit test, to ensure these do not interfere with rendered output

## Tested

* Verified in Chrome, Firefox, Safari on OSX
* Verified in IE11, Firefox, Chrome on Win7
* iPhone 8
* Galaxy 5, 6, 7, 8, 8+, 9, 9+ in Chrome and Firefox

### DIFF
--- a/src/components/navigation/Breadcrumbs/Breadcrumbs.unit.spec.jsx
+++ b/src/components/navigation/Breadcrumbs/Breadcrumbs.unit.spec.jsx
@@ -44,8 +44,7 @@ describe('<Breadcrumbs>', () => {
       <Breadcrumbs
         customClasses="foo test"
         id="foo"
-        listId="foo-list"
-        mobileWidth={375}>
+        listId="foo-list">
         {crumbs}
       </Breadcrumbs>
     );
@@ -56,7 +55,6 @@ describe('<Breadcrumbs>', () => {
     expect(navElem.props().className).to.include('va-nav-breadcrumbs');
     expect(navElem.props().className).to.include('foo test');
     expect(navElem.props().id).to.equal('foo');
-    expect(navElem.props()['data-mobile-width']).to.equal(375);
     expect(listElem.props().id).to.equal('foo-list');
   });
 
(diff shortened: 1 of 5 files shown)


src/components/navigation/Breadcrumbs/Breadcrumbs.unit.spec.jsx
```diff
@@ -163,7 +153,7 @@ describe('<Breadcrumbs>', () => {
   it('should render individual children correctly', () => {
     const tree = shallow(
       <Breadcrumbs>
-        <a href="/" key="1">Link 1</a>
+        <a href="/" onClick={() => { const result = {}; return result; }} key="1">Link 1</a>
```
1Copenut: @ncksllvn I added a basic `onClick` here to test rendering. Please advise if I need to remove it for security or bad practice considerations.

ncksllvn: No this makes perfect sense within React. As flat HTML, that's when the Metalsmith Nonce plugin takes effect and makes things kinda weird.

PR #135, opened 2018-07-19
Fix links

phiden:


jbalboni:
I think there are broken images in the content that's being added (anything starting with `/img` instead of `/design-system/img`).

PR #137, opened 2018-07-20
Subway maps

phiden:
@1Copenut I've still not been able to get the CSS to behave and I'm wondering if it's a build issue, so I'd like to test it on prod with this PR.


src/components/SubwayMap/SubwayMap.njk
```diff
@@ -0,0 +1,33 @@
+<div className="process schemaform-process">
```
1Copenut: @phiden These `className="classes..."` declarations can be changed to `class="classes..."` and the CSS will work as expected.

1Copenut:
This subway map is a great way to visually and aurally illustrate a stepped process. It has a lot of good navigational hooks for screen readers, and clearly delineates each step. Spot checking in a couple of screen readers (VoiceOver and NVDA) to ensure consistent announced behavior.

1Copenut:
Subway map good to go in VoiceOver and NVDA screen readers. Markup is sound and should support JAWS with no issues. I can also spot check in production if you have an example or URL you'd like me to review.

PR #138, opened 2018-07-24
Alert box margins

bshyong:
Prevents text overlap with alert box close button

Resolves https://github.com/department-of-veterans-affairs/vets.gov-team/issues/11334


BBarone1:
may also resolve: 
Fix formatting for message boxes on windows (likely site-wide issue) #11699

PR #140, opened 2018-07-26
Update Modal screen reader hint

erikphansen:
Switch to using an `aria-label` directly on the button element.
This seems to be a more modern, standards-compliant way to add
screen reader hints and fixes https://github.com/department-of-veterans-affairs/vets.gov-team/issues/11698

### DIFF
--- a/src/sass/modules/_m-overlay.scss
+++ b/src/sass/modules/_m-overlay.scss
@@ -7,21 +7,6 @@
 // See vets.gov-team#1047
 //=============================================================
 
-// TODO: Safe to delete this now? Nothing seems to actually have the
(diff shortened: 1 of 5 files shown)


src/sass/modules/_m-overlay.scss
```diff
@@ -4,9 +4,11 @@
 //
 // When the differences between our modals are
 // squared-away, we can remove this chunk.
-// See vets.gov-team#1047 
+// See vets.gov-team#1047
 //=============================================================
 
+// TODO: Safe to delete this now? Nothing seems to actually have the
+// `.va-overlay-close--icon` class applied to it.
```
rianfowler: 👍

PR #141, opened 2018-07-30
add click to close

rianfowler:
Verified with yalc

### DIFF
--- a/src/components/modal/Modal.jsx
+++ b/src/components/modal/Modal.jsx
@@ -14,8 +14,20 @@ function focusListener(selector) {
         focusableElement.focus();
       }
     }
+
   };
   document.addEventListener('focus', listener, true);
+  return listener;
+}
+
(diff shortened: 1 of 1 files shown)


src/components/modal/Modal.jsx
```diff
@@ -187,6 +211,7 @@ Modal.propTypes = {
 };
 
 Modal.defaultProps = {
+  clickToClose: false,
```
rianfowler: Don't want to change / test existing usages of the modal so I put this behavior behind an option flag.

src/components/modal/Modal.jsx
```diff
@@ -136,6 +156,10 @@ Modal.propTypes = {
    * Handler for when the modal is closed
    */
   onClose: PropTypes.func.isRequired,
+  /**
+   * Click outside modal will call onClose prop
```
rianfowler: Summarizes change

PR #143, opened 2018-08-02
11846 accessibility issues

rianfowler:
validated locally with yalc

### DIFF
--- a/src/components/navigation/Pagination/Pagination.jsx
+++ b/src/components/navigation/Pagination/Pagination.jsx
@@ -17,7 +17,11 @@ class Pagination extends React.Component {
     let nextPage;
     if (this.props.pages > this.props.page) {
       nextPage = (
-        <a aria-label="Next page" onClick={() => {this.props.onPageSelect(this.props.page + 1);}}>
+        <a
+          aria-label="Next page"
+          onClick={() => {this.props.onPageSelect(this.props.page + 1);}}
+          onKeyDown={e => this.handleKeyDown(e, this.props.page + 1)}
+          tabIndex="0">
           Next
         </a>
       );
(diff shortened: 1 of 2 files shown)


src/components/navigation/Pagination/Pagination.jsx
```diff
@@ -114,7 +129,9 @@ class Pagination extends React.Component {
           key={pageNumber}
           className={pageClass}
           aria-label={`Page ${pageNumber}`}
-          onClick={() => this.props.onPageSelect(pageNumber)}>
+          onClick={() => this.props.onPageSelect(pageNumber)}
+          onKeyDown={e => this.handleKeyDown(e, pageNumber)}
+          tabIndex="0">
```
rianfowler: Adds element to tab order.

src/components/navigation/Pagination/Pagination.jsx
```diff
@@ -114,7 +129,9 @@ class Pagination extends React.Component {
           key={pageNumber}
           className={pageClass}
           aria-label={`Page ${pageNumber}`}
-          onClick={() => this.props.onPageSelect(pageNumber)}>
+          onClick={() => this.props.onPageSelect(pageNumber)}
+          onKeyDown={e => this.handleKeyDown(e, pageNumber)}
```
rianfowler: I created a ticket to convert these to buttons so that we don't need a keydown handler. There's some styling refactoring that makes it difficult to handle in current sprint.

src/components/navigation/Pagination/Pagination.jsx
```diff
@@ -99,6 +107,13 @@ class Pagination extends React.Component {
     return range(start, end);
   }
 
+  handleKeyDown = (e, pageNumber) => {
+    if ((e.which || e.keyCode) === 13) {
```
cehsu: Should we also be listening for [space button keydown](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_button_role#Keyboard_and_focus)?

rianfowler: Yes- I just checked another button. I will add.

PR #151, opened 2018-08-27
Bc megamenu mobile fix

accbjt:
This is a fix for the MegaMenu scrolling issue on Mobile. IOS uses the resize event when scrolling so it causes the menu to hide. We don't want this. This checks if the size of the screen ever changes and if it doesn't it won't hide. You can only test this on an IOS device or on a simulator. Testing on Chrome won't be valid.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation",
-  "version": "1.4.5",
+  "version": "1.4.6",
   "license": "CC0-1.0",
   "repository": {
     "type": "git",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation",
-  "version": "1.4.1",
+  "version": "1.4.6",
```
ncksllvn: Why is this upgrade necessary?

accbjt: DOn't we need to update the patch number when we do a fix? The reason why it went from 1.4.1 to 1.4.6 is I was publishing the LeftNav stuff without merging to master. Next time I won't do that. I'll just leave development in vets-website and port it over once we have things ready.

ncksllvn: Gotcha! Makes sense.

PR #152, opened 2018-08-28
Adding props for focusable, role to header SVGs

1Copenut:
## Issue Being Solved

The Vets Website headers have a magnifying glass for the Search, and a question mark for the Help menus respectively. These cause issue in IE11 + JAWS because the SVG's can be focused inside the link, causing link text to be read twice.

We are solving for this issue, and on recommendation from @ncksllvn adding an `<IconBase />` component that gives a default `focusable="false` and a hook for setting `role` on the rendered `<SVG>`

## Added
* New `<IconBase />` to wrap current and future SVG icons
* Props `focusable:Boolean` and `role:string` to the `<IconBase />`
* Passed props into the Formation examples

## Tested with screen readers
* Win10 + IE11 + JAWS
* Win10 + IE11 + Chrome
* OSX + Safari + VoiceOver
* OSX + Chrome+ VoiceOver
* Win10 + Firefox + NVDA through the power of `yalc` on vets-website (localhost)


src/components/svgicons/IconHelp/IconHelp.jsx
```diff
@@ -4,10 +4,14 @@ import React from 'react';
 class IconHelp extends React.Component {
   render() {
     return (
-      <svg width="62" height="63"
-        viewBox="0 0 62 63"
+      <svg
         className={this.props.cssClass}
-        id={this.props.id}>
+        focusable={this.props.focusable}
```
cvalarida: If this prop isn't passed, this will be `undefined`, which is `falsy`. Does it make sense to make it non-focusable by default?

src/components/header/SearchMenu/SearchMenu.jsx
```diff
@@ -42,7 +42,7 @@ class SearchMenu extends React.Component {
             name="query"
             type="text"/>
           <button type="submit">
-            <IconSearch color="#fff"/>
+            <IconSearch color="#fff" focusable={false} role="presentation"/>
```
1Copenut: I forgot this on the first pass. There are two Search SVG icons:

1. The search icon that's `role="presentation"` next to "Search" string
2. The search icon `<button>` that has hidden "Search" text for screen readers

I retested for accuracy and it's 💯 on all screen readers.

src/components/header/HelpMenu/HelpMenu.jsx
```diff
@@ -6,7 +6,7 @@ import IconHelp from '../../svgicons/IconHelp/IconHelp';
 
 class HelpMenu extends React.Component {
   render() {
-    const icon = <IconHelp color="#fff"/>;
+    const icon = <IconHelp color="#fff" focusable={false} role="presentation"/>;
```
ncksllvn: If this is common behavior for SVGs, I wonder if there's a base SVG component where we can apply the `focusable` and `role` attributes.

src/components/svgicons/IconBase/IconBase.jsx
```diff
@@ -0,0 +1,32 @@
...
+IconBase.PropTypes = {
+  focusable: PropTypes.bool.isRequired,
+  role: PropTypes.string,
+};
+
+IconBase.defaultProps = {
+  focusable: false
+};
+
+export default IconBase;
```
ncksllvn: Very cool! Love it.

PR #153, opened 2018-08-30
Enable gold focus indicators on radio buttons

erikphansen:


src/sass/base/focus.scss
```diff
@@ -1,9 +1,3 @@
-.form-radio-buttons {
```
erikphansen: I don't know the reason this was added in the first place. There clearly must have been a reason to explicitly remove the focus indicators from radio buttons. If we do need to disable the focus indicators on radio buttons, in some cases, it should probably be done in a more precise manner.

cehsu: IIRC this was done for aesthetic consistency. Given the accessibility considerations, we should revisit this decision.

cehsu:
Could you please add screenshots?

cehsu:
Looping in @emilywaggoner as well for the design perspective on this.

accbjt:
Just for some context here is the original issue for the change. There are mockups in the issue that you may reference.

https://github.com/department-of-veterans-affairs/vets.gov-team/issues/6082

cehsu:
Looping in @juliaelman for additional design system perspective/awareness.

PR #156, opened 2018-09-04
Remove button focus style override

cehsu:
These changes allow the button focus style to be applied to buttons on form review pages.

screen shot 2018-09-04 at 1 52 33 pm
.

### DIFF
--- a/src/components/form/controls/ProgressButton/ProgressButton.jsx
+++ b/src/components/form/controls/ProgressButton/ProgressButton.jsx
@@ -17,7 +17,6 @@ class ProgressButton extends React.Component {
 
     return (
       <button
-        form="myform"
         type={this.props.submitButton ? 'submit' : 'button'}
         disabled={this.props.disabled}
         className={`${this.props.buttonClass} ${this.props.disabled ? 'usa-button-disabled' : null}`}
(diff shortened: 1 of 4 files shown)


erikphansen:
I assume that's the "after" screenshot? Before it was just the browser's default focus indicator?

src/sass/modules/_m-form-process.scss
```diff
@@ -78,9 +78,6 @@
     &:active {
       background-color: $color-primary-darkest;
     }
-    &:focus {
```
erikphansen: well that was easy!

src/components/form/controls/ProgressButton/ProgressButton.jsx
```diff
@@ -17,6 +17,7 @@ class ProgressButton extends React.Component {
 
     return (
       <button
+        form="myform"
```
erikphansen: The name of the form is always "myform"?

erikphansen:
Do you want to add the version bump to this PR or do that as a follow-up?

src/components/form/controls/ProgressButton/ProgressButton.jsx
cehsu: Thanks for asking, this change doesn't belong in this PR. Removed.

cehsu:
Thanks for the reminder @erikphansen, version bumped.

PR #158, opened 2018-09-05
Add document explaing how to make changes and view them locally

erikphansen:
This is a major rework of the instructions on making changes to the module and previewing them locally. I hope I made things cleared and more complete. I did remove what I thought were some unnecessary details in the old instructions. I also added info on tearing down the test set up and restoring things. The original instructions totally lacked those.

I'd encourage you to go through these instructions yourself and make sure that they are clear, accurate, and complete. If you do, I'd suggest that you simply change the [main background color](https://github.com/department-of-veterans-affairs/design-system/blob/f375a693d763888788d6cb2f9418ac85d4f8cdab/src/sass/base/_va.scss#L13) as a simple test to make sure you can preview the change in `vets-website`.

A few thoughts/questions:
- Why is this repo named `design-system` but the published module is `formation`? Not a big deal, but it can potentially lead to a little confusion.
- It looks like we could adjust the `test-publish` NPM script to first run `export-components`. It's hard to imagine a scenario where you'd want to publish locally without first building the module. Or maybe I should just make a new script that does both; it's silly to have to run two `design-system` scripts each time you want to see your changes.

### DIFF
--- a/previewing-changes.md
+++ b/previewing-changes.md
@@ -31,7 +31,7 @@ When you are done testing `formation` locally, you'll want to get your consuming
 
 In the consuming app:
 1. run `yalc remove @department-of-veterans-affairs/formation` to remove the link from the consuming app's `node_modules` dir to the local version of `formation`.
-2. **NOTE:** There seems to be an issue with Yalc not cleaning up after itself properly, so you'll have to also run `unlink node_modules/@department-of-veterans-affairs/formation`
+2. **NOTE:** There seems to be [an issue](https://github.com/whitecolor/yalc/issues/37) with Yalc not cleaning up after itself properly, so you'll have to also run `unlink node_modules/@department-of-veterans-affairs/formation`
 3. Run `yarn install --check-files` to reinstall `formation` from NPM. (If you're curious about why you need to `--check-files` flag, [check out this issue](https://github.com/yarnpkg/yarn/issues/2240).)
 
 Your consuming app should be back to normal!
--- a/previewing-changes.md
+++ b/previewing-changes.md
@@ -1,0 +1,37 @@
+# Making changes to design-system/formation and previewing them
+As you work in the `design-system` repo, you'll want to preview how your changes look in an app that uses `formation`. To do this, we use [Yalc](https://github.com/whitecolor/yalc) to publish `formation` locally, rather than to NPM, and tell the consuming app to install that local version of `formation`. In these examples, for the sake of simplicity, we'll assume you are using `vets-website` as the consuming app. But instructions should be similar for any app using `formation`.
+
+## Getting set up
+First, install Yalc globally if you haven't already: `yarn global add yalc`.
+
+While in the `design-system` app:
+1. Run `npm run export-components` to build the module into the `dist/formation/` dir.
+2. Run `npm run test-publish` to publish the module to your local `~/.yalc` dir.
+
+Then in the consuming app (`vets-website`):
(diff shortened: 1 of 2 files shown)


previewing-changes.md
```diff
@@ -0,0 +1,37 @@
...
+While in the `design-system` app:
+1. Make some changes to the code.
+2. Build the module and publish it locally to `~/.yalc` by running `npm run export-components` followed by `npm run test-publish`.
+
+Then in the consuming app:
+1. Run `yalc update` to bring in the latest version of `formation`.
+2. If you are using `vets-website` as your consuming app and it is already running, it should notice the changes and rebuild. If not, restart the app locally.
+3. After rebuilding completes, your changes will now be visible locally at `localhost:3001`.
+
+## Tearing down the local testing setup
```
cehsu: This is a great addition, good idea.

previewing-changes.md
```diff
@@ -0,0 +1,37 @@
...
+When you are done testing `formation` locally, you'll want to get your consuming app back to normal; that is, tell it to stop using the version of `formation` published to `~/.yalc` and go back to using the version from NPM.
+
+In the consuming app:
+1. run `yalc remove @department-of-veterans-affairs/formation` to remove the link from the consuming app's `node_modules` dir to the local version of `formation`.
+2. **NOTE:** There seems to be an issue with Yalc not cleaning up after itself properly, so you'll have to also run `unlink node_modules/@department-of-veterans-affairs/formation`
```
cehsu: Is it possible to provide a link to any bug reports here?

erikphansen: Sadly I didn't see an open issue for this (which means I should file one myself!). I also tested `yalc add` instead of `yalc link` to see if the cleanup went more smoothly. Sadly it did not.

I'll file a bug a add a link to it here.

erikphansen:
I just spent a chunk of time seeing it plain old `yarn link` would work better than `yalc`. It works, but because of how design-system builds its module, it's a lot easier to stick with `yalc` and deal with the fact that it doesn't clean up after itself very well. It's a minor annoyance that's easy to deal with as long as it's well documented.

previewing-changes.md
```diff
@@ -0,0 +1,37 @@
+# Making changes to design-system/formation and previewing them
```
juliaelman: Is there a reason why we use `yalk` vs. `npm link` when running locally? I have not used Yalc and am not entirely clear what the benefits are.

erikphansen: I'm not sure why `yalc` was chosen either. What I saw when doing [my experiment](https://github.com/department-of-veterans-affairs/design-system/pull/158#issuecomment-418846304) with `yarn link` was that whenever you triggered a rebuild of `design-system`, that would immediately cause `vets-website` to rebuild and blow up because either: (a) the package.json was no longer there, or (b) not all of the required module files were there. It's probably (a). So we could probably use `yarn watch`, but it would certainly require some changes to the build scripts in `design-system` and right now it doesn't seem worth that effort. `yalc` works nicely for the actual build-test loop, it just takes a little extra effort to clean up when you tear things down.

What I described above might be why `yalc` was chosen. I'd never used it before either. In our situation the fact that it takes an explicit command (`yalc update`) to tell the consuming app to get the new build is a benefit.

jbalboni: `yalc` was something @rianfowler implemented, he would be the one with the most background on it. But one of the benefits of `yalc` is that it doesn't use symlinks, which often have issues for Windows users. The disadvantages of `link` are summarized here: https://github.com/yarnpkg/yarn/issues/1761#issuecomment-259706202.

juliaelman: @jbalboni that sounds like a great benefit, especially when creating a distributed platform like Formation. TIL.

PR #159, opened 2018-09-06
Support optional aria label specifier for improved accessibility

cehsu:
These changes allow an optional aria-label specifier to be used in the pagination component's links.

### DIFF
--- a/src/components/navigation/Pagination/Pagination.jsx
+++ b/src/components/navigation/Pagination/Pagination.jsx
@@ -18,7 +18,7 @@ class Pagination extends React.Component {
     if (this.props.pages > this.props.page) {
       nextPage = (
         <a
-          aria-label="Next page"
+          aria-label={`Load next page ${this.props.ariaLabelSpecifier}`}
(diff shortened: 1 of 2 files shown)


cehsu:
Good idea, @erikphansen. Does "results" seem generic enough as a default specifier?

src/components/navigation/Pagination/Pagination.jsx
```diff
@@ -130,7 +130,7 @@ class Pagination extends React.Component {
         <a
           key={pageNumber}
           className={pageClass}
-          aria-label={`Page ${pageNumber}`}
+          aria-label={`Load page ${pageNumber} ${this.props.ariaLabelSpecifier}`}
```
rianfowler: suggest something like `ariaLabelSuffix` that specifies how this prop is used.

erikphansen:
It might be safer to skip using a default value; perhaps there's a case where "results" wouldn't make sense. Leaving it off would just result in the current behavior, so not really a regression.

PR #160, opened 2018-09-06
Add role="button" rule

cvalarida:
## Description

The gold focus ring wasn't being applied to the file uploader's button because it's not a _real_ button. It's a label masquerading as a button with a `[role="button"]`. This PR fixes that by applying the same style to anything that's acting like a button with the role attribute.

I'm not sure what all is needed before we can use this in vets-website, but I think I might need to build? `npm run pre-publish`? Does that need to happen before we can merge this into master, or is that only needed before we can publish to npm?

### DIFF
--- a/src/sass/base/focus.scss
+++ b/src/sass/base/focus.scss
@@ -11,7 +11,7 @@ select:focus {
   @include focus-gold-light-outline;
 }
 
-[role="button"], button:focus {
+[role="button"]:focus, button:focus {
   @include focus-gold-light-outline;
   position: relative;
   z-index: 3;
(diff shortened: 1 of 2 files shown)


src/sass/base/focus.scss
```diff
@@ -11,7 +11,7 @@ select:focus {
   @include focus-gold-light-outline;
 }
 
-button:focus {
+[role="button"], button:focus {
```
jbalboni: I think this needs to be `[role="button"]:focus`

jbalboni:
@cvalarida I think you also need to bump the version in package.json. Otherwise, you only need to run the publish commands when pushing to npm.

src/sass/base/focus.scss
cvalarida: :man_facepalming: Good catch. I didn't go through the whole `npm link` process because I tried it out in my element inspector. :roll_eyes:

PR #162, opened 2018-09-07
Add optional label to ErrorableCheckbox

erikphansen:


juliaelman:
🤔It's not 100% clear on what the interaction steps are for this form. To clarify, here are the steps *I think* a user experiences when interacting with this form:

1. User is on a page to search for a school
1. User types in school name in search box and hits submit
1. User views a list of schools and selects their school OR does not see their school

If this is accurate, this check box might only be needed once they have seen that list of search results. Is that correct?

erikphansen:
Thanks @juliaelman . @rianfowler might want to weigh in on that since he was more intimately involved with laying this all out and I know it went through some iterations.

src/components/form/controls/ErrorableCheckbox/ErrorableCheckbox.jsx
```diff
@@ -45,6 +45,9 @@ class ErrorableCheckbox extends React.Component {
 
     return (
       <div className={className}>
+        {this.props.labelAboveCheckbox && <span className="label-above-checkbox">
```
rianfowler: little concerned that we're calling this a label but not using the `<label>` - not sure what else to call it though.

cehsu: Perhaps subtitle instead?

rianfowler: or text or span. I think keeping generic views as literal as possible is good- it's conceptually a label but not in the context of markup.

erikphansen: This actually gives me an idea I'm gonna try out for this. It probably won't work.

erikphansen: Didn't work

PR #164, opened 2018-09-14
Add target attribute to megamenu for external links

RobAWilkinson:


src/components/navigation/MegaMenu/Column.jsx
```diff
@@ -61,6 +65,7 @@ Column.propTypes = {
     link: PropTypes.shape({
       text: PropTypes.string.isRequired,
       href: PropTypes.string.isRequired,
+      target: PropTypes.string,
```
ncksllvn: Do we need to have a default value applied to the `target` property?

RobAWilkinson: good call, I'll set it as `_self` in my tests if it was undefined, react just wasnt applying the attribute however

ncksllvn: Ah nice, totally slipped my mind that that's how React would handle it.

1Copenut:
Did we want to add a `rel="noopener noreferrer" declaration also? Thinking of this article when I see `_target`: https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/

RobAWilkinson:
I'd never seen this before @1Copenut  we have a ton of `target="_blanks"` without it scattered throughout the site.  

Since the pages that we are linking too are trusted VA content that we control, and not user-generated seems like it won't affect us.

PR #165, opened 2018-09-14
Optimize Lodash usage

jbalboni:
This adds the Babel Lodash plugin that converts imports to use the individual packages and also removes `lodash/fp` from some files.

This will let us reduce the size of the Lodash files included in our bundles on Vets.gov.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -63,6 +63,7 @@
       }],
     'no-plusplus': 0,
     'react/jsx-indent-props': 0,
+    'no-restricted-imports': ['error', 'lodash/fp'],
 
     // TODO: evaluate and potentially turn these rules back on
     'import/extensions': 0,


src/components/AdditionalInfo/AdditionalInfo.jsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import _ from 'lodash/fp';
+import _ from 'lodash';
```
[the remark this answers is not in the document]
jbalboni: I don't see any others when searching (there are still instances of regular lodash, though). The argument order didn't need to change in the places I changed, as far as I could tell.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation",
-  "version": "1.4.13",
+  "version": "1.4.14",
```
erikphansen: Could you update the version number in the `module-readme.md` as well?

src/components/form/controls/ErrorableDate/ErrorableDate.jsx
```diff
@@ -31,9 +30,13 @@ class ErrorableDate extends React.Component {
...
 
   handleChange(path, update) {
-    let date = set(path, update, this.props.date);
+    const date = {
+      ...this.props.date,
+      [path]: update
+    };
+
     if (!date.month.value) {
-      date = set('day.value', '', date);
```
erikphansen: Well this old code was an interesting abuse of the the FP `set` method :)

PR #166, opened 2018-09-17
Update required fields to match forms

jbalboni:
The required fields format for components in here do not match forms on vets-website, which are much more clear. In Formation, we're still using asterisks, but in vets-website we use `(*Required)`. Default USWDS uses `(required)`, I believe. This updates Formation to use the vets-website format.

screen shot 2018-09-17 at 2 32 37 pm
screen shot 2018-09-17 at 2 32 19 pm


jbalboni:
@cehsu This changes matches us-forms-system.

cehsu:
The provided link illustrates a case of the asterisk-only required label, should we set up a ticket to convert that as well?

jbalboni:
@cehsu Do you mean the linked issue? The asterisk in that issue will be updated when we pull in this change.

cehsu:
Thanks for clarifying @jbalboni.

PR #168, opened 2018-09-17
Collapsible panel button mg

markgreenburg:
- Adds a `type="button"` property to CollapsiblePanels so that they don't automatically submit forms on click.
- Removes extra `version` property in `package.json`
- Updates node modules readme with new version number


package.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation",
-  "version": "1.5.0",
-  "version": "1.4.14",
```
markgreenburg: 🤷‍♂️

PR #174, opened 2018-09-30
Mega Menu - Add onClick handler to links for event-tracking

ncksllvn:
This pull request is related to the work being done in https://github.com/department-of-veterans-affairs/vets-website/pull/8598.

We need a method of tracking when links are clicked, and the best way of doing that is by adding the `onClick` events to the links so that we can capture the event before the page navigates.

### DIFF
--- a/module-readme.md
+++ b/module-readme.md
@@ -1,4 +1,4 @@
-Version 1.7.1
+Version 1.7.2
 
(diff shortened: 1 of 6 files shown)


jbalboni:
@ncksllvn Are you looking to release this soon? May want to add a version bump to make that process smoother.

ncksllvn:
@jbalboni Ah yes forgot! Updated the patch version!

PR #176, opened 2018-10-01
Megamenu - Add analytics into a couple missing places

ncksllvn:
I got ahead of myself and merged https://github.com/department-of-veterans-affairs/design-system/pull/174#event-1877396390 while missing a couple of places where the `linkClicked` needed to be passed.

### DIFF
--- a/src/components/navigation/MegaMenu/MegaMenu.jsx
+++ b/src/components/navigation/MegaMenu/MegaMenu.jsx
@@ -67,7 +67,9 @@ export default class MegaMenu extends React.Component {
             defaultSection={defaultSection(item.menuSections)}
             currentSection={currentSection}
             updateCurrentSection={() => this.updateCurrentSection(section.title)}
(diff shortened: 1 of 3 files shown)


src/components/navigation/MegaMenu/MegaMenu.jsx
```diff
@@ -77,7 +79,9 @@ export default class MegaMenu extends React.Component {
         data={item.menuSections}
         navTitle={item.title}
         handleBackToMenu={() => this.toggleDropDown('')}
-        show={this.props.currentDropdown !== ''}></SubMenu>
+        show={this.props.currentDropdown !== ''}
+        linkClicked={this.props.linkClicked}
+        columnThreeLinkClicked={this.props.columnThreeLinkClicked}></SubMenu>
```
ncksllvn: Ideally, this component would only be instantiated once, but until we have more bandwidth this is the easiest solution.

PR #178, opened 2018-10-02
Don't change button position on focus

cvalarida:
We had a problem of bouncing the buttons around when clicked because the `position` changed.

This PR stops the position from changing once clicked.

I don't know why it was in here in the first place, so if you know of why it was needed originally, please let me know!

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation",
-  "version": "1.7.3",
+  "version": "1.7.4",
   "license": "CC0-1.0",
   "repository": {
     "type": "git",
--- a/src/sass/base/focus.scss
+++ b/src/sass/base/focus.scss
@@ -13,7 +13,6 @@ select:focus {
 
 [role="button"]:focus, button:focus {
   @include focus-gold-light-outline;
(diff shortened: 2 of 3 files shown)


src/sass/base/focus.scss
```diff
@@ -13,7 +13,6 @@ select:focus {
 
 [role="button"]:focus, button:focus {
   @include focus-gold-light-outline;
-  position: relative;
   z-index: 3;
```
jbalboni: I wonder why we have this, too.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation",
-  "version": "1.7.3",
+  "version": "1.7.4",
```
jbalboni: There's also a version in module-readme.md

markgreenburg:
Tracked the changes to [this PR](https://github.com/department-of-veterans-affairs/design-system/pull/123) which links to [this ticket](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/6082). Also links to corresponding `vets-website` issue [here](https://github.com/department-of-veterans-affairs/vets-website/pull/7676).

PR #183, opened 2018-10-16
Reset background-color for accordion buttons on active

jbalboni:
This overrides the standard button active style for accordion buttons, which are grey backgrounded by default.

screen shot 2018-10-16 at 3 41 13 pm


jbalboni:
I will probably release this with the menu update, but might as well update it to be safe.

PR #188, opened 2018-10-23
Use embedded svg

rianfowler:
mega menu uses a relative path to an image that is not available on team sites. opted to embed the svg directly into the component. 

screen shot 2018-10-22 at 17 47 21


src/components/navigation/MegaMenu/SubMenu.jsx
```diff
@@ -52,7 +53,7 @@ const SubMenu = ({ data, show, navTitle, handleBackToMenu, linkClicked, columnTh
           seeAllLink && <div className="panel-bottom-link">
             <a href={seeAllLink.href} onClick={linkClicked.bind(null, seeAllLink)}>
               View All in {seeAllLink.text}
-              <img className="all-link-arrow" src="/img/arrow-right-blue.svg" alt="right-arrow"></img>
+              <ArrowRightBlueSVG/>
```
rianfowler: This is one pattern we can follow for portable components.

src/components/navigation/MegaMenu/arrow-right-blue.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+export const ArrowRightBlueSVG = () => (
+  <svg
+    width="444.819"
```
erikphansen: Roughly 444.819px wide :-D

src/components/navigation/MegaMenu/SubMenu.jsx
erikphansen: This works for me but would love to hear thoughts from other people regarding this approach.

rianfowler: I think there are some other ways to deal with svgs + react + webpack. I think we need to get a handle on how we want people to used this stuff e.g. do we want them to build it and host their own assets or do we want them to be able to just throw a script on a page.

PR #190, opened 2018-10-25
Brings hub page link-list in as component with example

RobAWilkinson:


src/sass/modules/_m-hub-page-link-list.scss
```diff
@@ -0,0 +1,34 @@
+.hub-page-link-list {
+  // override default ul styles
+  list-style: none;
+  padding: 0;
+  .hub-page-link-list--item {
```
jbalboni: Does this need to be nested under the `.hub-page-link-list` selector?

Also, I think we should avoid using `--` in these classes unless we're trying to use BEM (in which case, `item` is an element and it would be `hub-page-link-list__item`). Otherwise, I think we confuse people who do know BEM (and wonder why it's wrong) and also those who don't (who would wonder why you would use `--`).

Though, I could be completely wrong and `--` is a more common separator than just in BEM.

src/components/HubPageLinkList/HubPageLinkList.njk
```diff
@@ -0,0 +1,18 @@
+<h2 class="hub-page-link-list--title">
```
jbalboni: Eventually It'd be nice to get things sorted out so we could import this template directly in vets-website, instead of having to duplicate it. I haven't actually tried that, but I assume it wouldn't work because of the slightly different template engines.

RobAWilkinson: I could easily make it a template in vets-website @jbalboni, let me give it a try

src/sass/modules/_m-hub-page-link-list.scss
```diff
@@ -0,0 +1,34 @@
+.hub-page-link-list {
+  // override default ul styles
+  list-style: none;
+  padding: 0;
+  .hub-page-link-list__item {
```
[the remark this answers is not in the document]
RobAWilkinson: wow thanks for the tip @jbalboni  I've always used that for `&:hover` but never for just plain strings.
Exactly I hate nesting super specific css classes like this, kinda defeats the purpose, if sass converts it to a top level, than its just syntactic sugar.

src/components/HubPageLinkList/HubPageLinkList.config.yml
```diff
@@ -0,0 +1,63 @@
...
+        external: false
+      - url: https://www.cem.va.gov/cems/listcem.asp
+        label: Find a National, State, or Tribal Government Veterans Cemetery
+        description: Search for national, state, or tribal government Veterans cemeteries.
+        external: false
+      - url: "/records/get-military-service-records/"
+        label: Request Military Records (DD214)
+        description: View and print documents from your official military personnel file
+          (including your DD214).
+        external: false
```
ncksllvn: Solid example!

src/sass/modules/_m-hub-page-link-list.scss
```diff
@@ -0,0 +1,42 @@
...
+  }
+}
+.external-link-icon-black {
+  background-image: url(/img/icons/exit-icon.png);
+  background-position: 100% 50%;
+  background-repeat: no-repeat;
+  background-size: 1em auto;
+  padding-right: 1.2em;
+  position: relative;
+}
```
ncksllvn: Since this class is kinda a utility or general use, I wonder if there’s a more discoverable place it could live.

PR #194, opened 2018-10-30
Fixed no background image alerts

U-DON:
Follow-up to #193. Forgot to account for the "alerts" that had no icons or borders.

I've renamed the class from `no-background-image` to `background-color-only` for accuracy. It was only being used for `usa-alert` elements, so I think the impact should be isolated. I will be updating `vets-website` and `vagov-content` accordingly.

Since the absolute positioning of the icon on the default alerts didn't seem necessary, I made some changes to that as well along with minor spacing tweaks.

screen shot 2018-10-29 at 9 50 57 pm

### DIFF
--- a/module-readme.md
+++ b/module-readme.md
@@ -1,4 +1,4 @@
-Version 1.10.0
+Version 1.10.1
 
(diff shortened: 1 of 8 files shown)


U-DON:
Yup, the designs for this variation will be revisited separately. Just keeping these consistent rather than making a sweeping change that forces all types of `usa-alert-*` to look like the new alert.

PR #195, opened 2018-10-30
Higher precedence rule for usa-alert-body.

U-DON:
Really minor change to override a `padding-left: 5rem` rule from USWDS.

### DIFF
--- a/src/sass/modules/_m-alert.scss
+++ b/src/sass/modules/_m-alert.scss
@@ -15,7 +15,8 @@
     position: static;
   }
 
-  &-body {
+  // &-body would not have had high enough precedence.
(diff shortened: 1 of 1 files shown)


src/sass/modules/_m-alert.scss
```diff
@@ -15,7 +15,8 @@
     position: static;
   }
 
-  &-body {
+  // &-body would not have had high enough precedence.
+  & .usa-alert-body {
```
jbalboni: I think you can ditch the `&` here.

U-DON: Good point!

PR #197, opened 2018-10-31
Use match media to test screen size

jbalboni:
So it turns out that using `document.body.clientWidth` doesn't account for the scrollbar that may appear on desktop pages. I didn't think this was an issue, but it leaves a gap in our media queries vs our code, leading to the linked issue.

Since `window.innerWidth` is flaky on device emulators, I thought a better option generally would be to use `matchMedia`.

screen shot 2018-10-31 at 2 24 38 pm

### DIFF
--- a/src/components/navigation/MegaMenu/MegaMenu.jsx
+++ b/src/components/navigation/MegaMenu/MegaMenu.jsx
@@ -20,7 +20,7 @@ export default class MegaMenu extends React.Component {
       this.props.toggleDisplayHidden(true);
     }
 
-    mobileMediaQuery.addListener(this.resetDefaultState.bind(this));
+    mobileMediaQuery.addListener(this.resetDefaultState);
     document.body.addEventListener('click', this.handleDocumentClick, false);
   }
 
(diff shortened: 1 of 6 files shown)


src/components/navigation/MegaMenu/MegaMenu.jsx
```diff
@@ -4,39 +4,36 @@ import MenuSection from './MenuSection';
...
-  }
-
   componentDidMount() {
-    if (document.body.clientWidth < 768) {
+    if (mobileMediaQuery.matches) {
       this.props.toggleDisplayHidden(true);
     }
 
-    window.addEventListener('resize', this.resetDefaultState.bind(this));
+    mobileMediaQuery.addListener(this.resetDefaultState.bind(this));
```
jbalboni: Conveniently, this only fires when the media query result changes, so far fewer events will trigger our callback.

src/components/navigation/MegaMenu/MegaMenu.jsx
```diff
@@ -4,39 +4,36 @@ import MenuSection from './MenuSection';
...
+    mobileMediaQuery.addListener(this.resetDefaultState);
     document.body.addEventListener('click', this.handleDocumentClick, false);
   }
 
   /**
    * Remove event listener
    */
   componentWillUnmount() {
-    window.removeEventListener('resize', this.resetDefaultState.bind(this));
+    mobileMediaQuery.removeListener(this.resetDefaultState);
```
jbalboni: I believe `bind` creates a new function, which means `removeEventListener` was never finding the actually attached callback to remove. This should fix that.

PR #200, opened 2018-11-12
Add heading classes

jbalboni:
USWDS styles heading elements directly, so we often pick a heading level based on style, instead of the markup structure of the page. This ends up causing accessibility issues.

This PR adds some heading classes so that we can flexibly style `h` elements and choose the correct one for the page.

### DIFF
--- a/src/sass/base/_b-utils.scss
+++ b/src/sass/base/_b-utils.scss
@@ -40,3 +40,25 @@
 .js-hide {
   display: none !important;
 }
+
+.heading-level {
+  &-1 {
+    @include h1();
+  }
+  &-2 {
+    @include h2();
+  }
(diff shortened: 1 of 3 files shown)


src/sass/base/_b-utils.scss
```diff
@@ -40,3 +40,25 @@
 .js-hide {
   display: none !important;
 }
+
+.heading-level {
```
jbalboni: We could just let people use the mixins, but I feel like this is more likely to be used.

1Copenut: If my Sass isn't too rusty, this would translate to classes `heading-level-1`, `heading-level-2`... Perfectly workable and understandable.

PR #207, opened 2018-12-03
Additional information heading

cehsu:
These changes update the additional information component to support setting the trigger text as an h4 element, using the `isHeading` prop.

image

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation",
-  "version": "1.12.0",
+  "version": "1.12.1",
   "license": "CC0-1.0",
   "repository": {
     "type": "git",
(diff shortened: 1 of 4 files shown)


src/components/AdditionalInfo/AdditionalInfo.jsx
```diff
@@ -30,10 +30,14 @@ export default class AdditionalInfo extends React.Component {
         aria-expanded={this.state.open ? 'true' : 'false'}
         aria-controls={this.expandedContentId}
         onClick={this.toggle}>
-        <span className="additional-info-title">
+        {this.props.isHeading && <h4 className="additional-info-title">
```
jbalboni: Could we do something like this to make the heading level configurable? I'm hesitant to force a heading level in this component.

https://stackoverflow.com/a/47124374

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation",
-  "version": "1.12.0",
+  "version": "1.12.1",
```
jbalboni: Also need to update this in module-readme.md

src/components/AdditionalInfo/AdditionalInfo.jsx
cehsu: Great idea, thanks @jbalboni!

PR #208, opened 2018-12-05
Apply focus to modals that are visible on mount

ncksllvn:
Modals that have the `visible` property set to `true` upon mount aren't receiving keyboard focus, as is visible on the VA.gov Welcome Modal. In the looking at the source for the Modal, there was more setup process involved for the modal than just the keyboard focus, so I consolidated that setup process into a single function, `setupModal`, and the teardown process into `teardownModal`. Then, we could just add a call to `setupModal` in `componentDidMount` when `this.props.visible` is true.

Related ticket - https://github.com/department-of-veterans-affairs/vets.gov-team/issues/14903

## Screenshots
This super grainy video refreshes the page, and shows the Welcome Modal open with focus. Then it's closed using the escape key.

modal-focus

## Testing done
- Tested the Welcome Modal
- Tested the edit modals on `/profile`

### DIFF
--- a/src/components/modal/Modal.unit.spec.jsx
+++ b/src/components/modal/Modal.unit.spec.jsx
@@ -10,6 +10,7 @@ describe('<Modal/>', () => {
   it('should render', () => {
     const tree = mount(<Modal id="modal" title="Modal title" visible onClose={() => {}}>Modal contents</Modal>);
     expect(tree.text()).to.contain('Modal contents');
+    tree.unmount();
   });
 
   it('should setup event listeners and tear them down', () => {
--- a/src/components/modal/Modal.unit.spec.jsx
+++ b/src/components/modal/Modal.unit.spec.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
-import { shallow } from 'enzyme';
+import sinon from 'sinon';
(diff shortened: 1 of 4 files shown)


src/components/modal/Modal.unit.spec.jsx
```diff
@@ -1,16 +1,33 @@
...
+import { mount } from 'enzyme';
 import { expect } from 'chai';
 import { axeCheck } from '../../../lib/testing/helpers';
 
 import Modal from './Modal.jsx';
 
 describe('<Modal/>', () => {
   it('should render', () => {
-    const tree = shallow(<Modal id="modal" title="Modal title" visible onClose={() => {}}>Modal contents</Modal>);
+    const tree = mount(<Modal id="modal" title="Modal title" visible onClose={() => {}}>Modal contents</Modal>);
```
ncksllvn: Switched to `mount` so that `ref`s work, https://github.com/airbnb/enzyme/issues/316.

rianfowler: [Recommend](https://github.com/department-of-veterans-affairs/vets.gov-team/pull/15504/files) running `tree.unmount()` at the end of this unit test.

PR #209, opened 2018-12-10
Margin utility

jaredcunha-usds:
# Utilities

Utilities will be single-property to add flexibility and/or override default CSS properties of components, and allow the creation of UI patterns without needing the write new CSS. The margin utility is the first of many, but also the largest.

## What's in the PR
- Sass maps for breakpoints and 8pt spacing units
- Margin utility CSS 

## How the margin utility works
The margin utility is very similar to the one [USWDS 2.0 Margin Utility](https://v2.designsystem.digital.gov/utilities/margin-and-padding/). Rather than using `0` as a separator for half values (2.5, 1.5, 0.5), we are using `p`, resulting in `vads-u-margin--1p5`. 

### Responsive margins
The margin utility can be altered by breakpoint with a `:` preceded by a named breakpoint that matches our current Sass variables for breakpoints. 
```
<div class="vads-u-margin--1 medium-screen:vads-u-margin--2 large-screen:vads-u-margin--4">
```

### DIFF
--- a/module-readme.md
+++ b/module-readme.md
@@ -1,4 +1,4 @@
-Version 1.12.2
+Version 1.13.0
 
 This module contains reusable react components from [vets-website](https://github.com/department-of-veterans-affairs/vets-website) housed in its design system [repo](https://github.com/department-of-veterans-affairs/design-system).
 
@@ -42,6 +42,10 @@ See [design system](https://department-of-veterans-affairs.github.io/design-syst
 - SystemDownView.js
 - Tooltip.js
 - MegaMenu.js
+
+## Included utilities:
+
+- Margins
 
 ## Styles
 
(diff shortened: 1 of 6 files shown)


jaredcunha-usds:
This also reflects a decision to move forward on the 8pt spacing system and naming conventions from no objections to the RFCs posted in the following issues: 
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/15436
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/15426

jaredcunha-usds:
@jbalboni - I'm second-guess if it's worth having a selector for negative margins on all sides. I can x and y, top, left, bottom, right.

jbalboni:
I guess thinking about it more, I would lean towards leaving them out. I don't find them useful very often.

PR #210, opened 2018-12-11
Publish assets and complete CSS file in Node module

jbalboni:
This PR builds a complete CSS file and includes it and other font and image assets in the npm module.


jaredcunha-usds:
Cool - if I'm reading through this, I might be able to edit the build script i have for the documentation site, right? If so, @jbalboni, might need to get in touch later this week if I run into difficulty.

jbalboni:
@jaredcunha-usds Yeah, that's the hope, at least. Ideally you would just use what's in the `dist` directory of the formation module.

PR #211, opened 2018-12-11
Modal - Check visible prop before applying focus

ncksllvn:
This is a follow-up to https://github.com/department-of-veterans-affairs/design-system/pull/208.

I noticed a console error in the `SignInModal` after it's closed. `handleDocumentFocus` was firing after closing the modal, and `this.element` of the modal instance was `null` because it was no longer visible. This is fixed in this PR by checking `this.props.visible` in the focus handler.


module-readme.md
```diff
@@ -1,4 +1,4 @@
-Version 1.14.0
+Version 1.14.1
```
ncksllvn: @jbalboni Is this the version I should be using? I noticed that this was out of sync with the one in the `package.json`.

jbalboni: Yeah, I think that's right.

PR #212, opened 2018-12-13
Visibility display

jaredcunha-usds:
This PR creates a visibility and display utility that can be used in place of the ones we have from Foundation. This implementation focuses on the CSS property rather than the concept, hence why it includes two utilities. 

Both utilities have responsive prefixes. 

This PR also includes a bug fix where the auto margin properties were outside the breakpoint loop.

### DIFF
--- a/src/sass/utilities/_visibility.scss
+++ b/src/sass/utilities/_visibility.scss
@@ -1,0 +1,43 @@
+.vads-u-visibility--hidden {
+  visibility: hidden !important;
+}
+
+.vads-u-visibility--visible {
+  visibility: visible !important;
+}
+
+.vads-u-visibility--screen-reader {
+  border: 0;
+  clip: rect(0, 0, 0, 0);
+  height: 1px;
+  overflow: hidden;
+  padding: 0;
+  position: absolute !important;
+  width: 1px;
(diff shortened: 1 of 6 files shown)


src/sass/utilities/_visibility.scss
```diff
@@ -0,0 +1,43 @@
+.vads-u-visibility--hidden {
+  visibility: hidden !important;
+}
+
+.vads-u-visibility--visible {
+  visibility: visible !important;
+}
+
+.vads-u-visibility--screen-reader {
```
jbalboni: Does this match up with the existing `usa-sr-only` class implementation?

jaredcunha-usds: `.usa-sr-only` uses only `position: absolute; left: -999em`.

Jonathan Snook has outlined an issue with that method:
> (Position Absolute and Offscreen) In this case, if you have focusable content within the positioned element, the page will scroll to that element, thus creating an unsettling jump for sighted users.

The way I have here uses a clip and nearly collapse. [source](https://snook.ca/archives/html_and_css/hiding-content-for-accessibility)

jaredcunha-usds:
Merged. @jbalboni Let's connect Friday to go through publishing the changes.

PR #219, opened 2019-01-09
[Sidenav] Fix for text falling on focus state

ncksllvn:
We had some additional style rules applied during the focus state of the side navigation list items. Those rules were added because the gold outline around the focused item was getting cut off. This PR refines that by using `outline-offset` instead of `margin` and `width`.

## Before
The "and" here is seen wrapping to the next line on focus -
image

## After
Here's the same item with the changes in this PR -
image

### DIFF
--- a/src/sass/modules/_m-nav-sidebar.scss
+++ b/src/sass/modules/_m-nav-sidebar.scss
@@ -152,6 +152,9 @@ $level-3-hover-padding: 8px 12px 8px 30px;
             border-color .1s ease-in-out 0s,
             padding .1s ease-in-out 0s;
         }
+        &:focus {
+          outline-offset: -2px;
+        }
       }
 
       i {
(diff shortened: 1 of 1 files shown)


src/sass/modules/_m-nav-sidebar.scss
```diff
@@ -312,22 +315,6 @@ $level-3-hover-padding: 8px 12px 8px 30px;
...
-        }
-      }
-    }
-
-    .usa-current {
-      // added to deal with accordian overflow
-      // and outline not showing correctly
-      &:focus {
-        width: 99% !important;
-        margin-top: 2px;
```
ncksllvn: These rules were accounting for the issue seen below -

image

ncksllvn:
@emilywaggoner I noticed this after the demo today. Is it okay that the outline is overlapping with the border as shown in the screenshots? This is seen while tabbing through the elements.

src/sass/modules/_m-nav-sidebar.scss
```diff
@@ -152,6 +152,9 @@ $level-3-hover-padding: 8px 12px 8px 30px;
             border-color .1s ease-in-out 0s,
             padding .1s ease-in-out 0s;
         }
+        &:focus {
+          outline-offset: -2px;
```
rianfowler: MDN [indicates](https://developer.mozilla.org/en-US/docs/Web/CSS/outline-offset) this isn't supported on IE.  Couldn't find any easy alternatives. [This](url
https://css-tricks.com/almanac/properties/o/outline-offset/) says it's partially supported with `outline` shorthand but I don't see how you can use that to achieve an offset.

jbalboni:
This is pretty far out of date now, I'm going to close it.

PR #221, opened 2019-01-14
Remove Fractal and split out React components

jbalboni:
This PR removes Fractal and reorganizes this repo as a monorepo with two modules:

- formation: this contains styles and assets for the design system and will include plain JS for some components in the future
- formation-react: this contains React implementations of some components in the design system

Issue: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/15406

In order to review this, it's easier to look at the branch directly and make comments in this PR that aren't attached to specific changes. There are too many changes otherwise.


packages/formation-react/scripts/build.js
```diff
@@ -52,29 +45,13 @@ fileNames.forEach(fileName => {
...
   // transform the buffer with babel using babelrc
   const babelTransformedBuffer = babel
-    .transform(fileBuffer, { 'extends': './.babelrc' })
+    .transform(fileBuffer, { 'extends': '../../.babelrc' })
     .code;
   // flatten paths given to all requires
   const requireFlattenedBuffer = flattenRequires(babelTransformedBuffer.toString());
   const newFileName = `${path.parse(fileName).name}.js`;
 
   // write file to dist/build
```
rianfowler: I don't think this comment is accurate anymore.

packages/formation-react/.npmignore
```diff
@@ -0,0 +1,3 @@
+src/*
```
rianfowler: using npmignore and gitignore is a big improvement

packages/formation-react/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@department-of-veterans-affairs/formation-react",
+  "version": "1.18.0",
```
rianfowler: should this be a major release since it's a breaking change?

jbalboni: Yes, I haven't updated versions yet because I think we'll want to change that process so that PRs don't ever touch version numbers and instead we use `lerna publish` on the master branch to tag releases in git and publish modules.

PR #223, opened 2019-01-15
a11y fixes for Modal component

erikphansen:


jbalboni:
@erikphansen Are you looking to release this immediately? If so, you should update the versions in package.json and module-readme.md.

PR #224, opened 2019-01-21
[MegaMenu] Increase height for new hub illustrations

ncksllvn:
Blocks https://github.com/department-of-veterans-affairs/vagov-content/pull/232

The new hub illustrations have a different aspect ratio than the former, causing text in the marketing containers to get cut off. See [here](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13822#issuecomment-456157255) for more context. This PR increases the height of the MegaMenu to account for that.

image

These illustrations are another part of upgrading Font Awesome, as the icons used in the PNGs are from FA.

### DIFF
--- a/module-readme.md
+++ b/module-readme.md
@@ -1,4 +1,4 @@
-Version 2.0.0
+Version 2.0.1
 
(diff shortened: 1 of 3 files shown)


ncksllvn:
Can someone take a look at this when they get the chance? We're trying to have this out by tomorrow.

PR #227, opened 2019-01-23
Font Awesome - Adds Brand, Regular icon sets

aubreyarcangel:
This PR adds .fab class to core.scss to include the brand icons in Font Awesome 5.

Note: I was not able to test this locally because of an earlier PR that needed to be merged first before my local instance. department-of-veterans-affairs/vets-website#9434

### DIFF
--- a/packages/formation/package.json
+++ b/packages/formation/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation",
-  "version": "3.0.0-FA-0.0.1",
+  "version": "3.0.0-FA-0.0.2",
   "description": "The VA design system",
   "keywords": [
     "va",
--- a/packages/formation/package.json
+++ b/packages/formation/package.json
@@ -24,11 +24,11 @@
     "url": "https://github.com/department-of-veterans-affairs/design-system/issues"
   },
   "dependencies": {
-    "@fortawesome/fontawesome-free": "^5.6.3",
-    "foundation-sites": "5",
     "moment": "^2.23.0"
   },
   "peerDependencies": {
+    "@fortawesome/fontawesome-free": "^5.6.3",
+    "foundation-sites": "5",
     "uswds": "^1.4.2"
   }
 }
--- a/packages/formation/package.json
+++ b/packages/formation/package.json
@@ -24,11 +24,11 @@
     "url": "https://github.com/department-of-veterans-affairs/design-system/issues"
   },
   "dependencies": {
-    "@fortawesome/fontawesome-free": "^5.6.3",
     "foundation-sites": "5",
     "moment": "^2.23.0"
   },
   "peerDependencies": {
+    "@fortawesome/fontawesome-free": "^5.6.3",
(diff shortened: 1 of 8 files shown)


ncksllvn:
I think this will have to wait until we can fully test this, reason being that I think we need to start importing the [Font Awesome brands.scss](https://github.com/FortAwesome/Font-Awesome/blob/master/scss/brands.scss). They use the `Font Awesome Brands` font family, but many (all?) are [free](https://fontawesome.com/icons?d=gallery&s=brands&m=free).

May also be worth expanding this to also import Font Awesome Regular, based on the icons in use in the prototype at https://github.com/department-of-veterans-affairs/vets.gov-team/issues/16242.

jbalboni:
FYI: nothing should be blocking testing this in vets-website now.

packages/formation/sass/core.scss
```diff
@@ -11,26 +11,12 @@
...
-
-.fa,
-.fas {
-  font-family: 'Font Awesome 5 Free';
-  font-weight: 900;
-}
+@import "~@fortawesome/fontawesome-free/scss/fontawesome.scss";
+@import "~@fortawesome/fontawesome-free/scss/solid";
+@import "~@fortawesome/fontawesome-free/scss/brands";
+@import "~@fortawesome/fontawesome-free/scss/regular";
```
ncksllvn: The issue was with the Webpack config in `vets-website` trying to inline the whole SVG. By changing that setting in the Webpack config, we can just import the FA SCSS no problem in the design system. See PR here, https://github.com/department-of-veterans-affairs/vets-website/pull/9451.

webpack.config.prod.js
```diff
@@ -39,7 +39,7 @@ module.exports = {
       {
         test: /\.svg/,
         use: {
-          loader: 'svg-url-loader'
+          loader: 'svg-url-loader?limit=1024'
```
ncksllvn: For consistency with https://github.com/department-of-veterans-affairs/vets-website/pull/9451/files#diff-a58d55bdb5770c78ad512f8e91f8d051R132

packages/formation/package.json
```diff
@@ -24,11 +24,11 @@
     "url": "https://github.com/department-of-veterans-affairs/design-system/issues"
   },
   "dependencies": {
-    "@fortawesome/fontawesome-free": "^5.6.3",
-    "foundation-sites": "5",
     "moment": "^2.23.0"
   },
   "peerDependencies": {
+    "@fortawesome/fontawesome-free": "^5.6.3",
+    "foundation-sites": "5",
```
ncksllvn: I moved these to peer dependencies because if the parent project wants to require our SCSS, then these do have to be installed alongside it.

ncksllvn:
https://github.com/department-of-veterans-affairs/vets-website/pull/9451 is set up to use the alpha version of this. Can you all take a look at the PR and approve if it looks okay? If so, I'm going to merge this, publish an official version, then update that PR to use the official version instead of alpha.

PR #228, opened 2019-01-24
Fix/make review button full width

rianfowler:
Match add another button full width on smaller screens. 

screen shot 2019-01-24 at 12 26 17
screen shot 2019-01-24 at 12 26 33
screen shot 2019-01-24 at 12 26 48
screen shot 2019-01-24 at 12 27 15

Targeted medium because the button wraps at that breakpoint on this form: 
screen shot 2019-01-24 at 12 28 03


packages/formation/sass/modules/_m-form-process.scss
```diff
@@ -65,6 +65,14 @@
 
 // Overriding the accordion button styles that are too broad
 .form-review-panel {
+  @include media-maxwidth($medium-screen) {
```
jbalboni: What is `media-maxwidth`? I've never seen that.

jbalboni: Oh, ok. I'm guessing this switches it from the usual `media` generated breakpoints which are `min-width` to using `max-width`

jbalboni:
@rianfowler I'm assuming this needs a version bump, too?

packages/formation/sass/modules/_m-form-process.scss
rianfowler: updated- behavior remains the same.

PR #229, opened 2019-01-24
Create Flex grid

jaredcunha-usds:
This will create a flexbox grid outlined in https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/vets.gov-team/16501

The flexbox grid will live with the USWDS grid, but it provides a little more flexibility due to the features available in flexbox. Once a flexbox utility is created, there will be an even wider suite of options.

More documentation on how the flexbox grid works, including responsive examples, can be found on the documentation site:
https://department-of-veterans-affairs.github.io/vets-design-system-documentation/layout/flexbox-grid.html

### DIFF
--- a/packages/formation-react/package.json
+++ b/packages/formation-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation-react",
-  "version": "3.0.0",
+  "version": "3.0.1",
   "description": "React implementation of patterns in Formation",
   "keywords": [
     "react",
--- a/packages/formation/package.json
+++ b/packages/formation/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation",
-  "version": "3.0.1",
+  "version": "3.1.0",
   "description": "The VA design system",
   "keywords": [
     "va",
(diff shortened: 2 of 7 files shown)


packages/formation-react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation-react",
-  "version": "3.0.0",
+  "version": "3.1.0",
```
jbalboni: You don't need to bump this one. I'm thinking we'll just try to keep the major version in sync, though we'll have to see how it feels over time.

jaredcunha-usds: That happened when I ran `npm run version`. Seems to bump both?

jbalboni: It should ask you for each individual module, but I guess I don't remember if it lets you skip updating one of the module.

jaredcunha-usds: ah, i actually ran `npm run version minor` which updated both. 

`npm run version` let me select… however, i did a custom version for `formation-react` to keep it the same. Is there not a _do not update_ choice?

Anyway, this should be good to go now.

PR #231, opened 2019-01-29
Flexbox utility

jaredcunha-usds:
This pull will create a Flexbox utility. This utility is in use and documented on the design.va.gov temporary site:
https://department-of-veterans-affairs.github.io/vets-design-system-documentation/utilities/flexbox.html

### DIFF
--- a/packages/formation/package.json
+++ b/packages/formation/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation",
-  "version": "3.2.0",
+  "version": "3.3.0",
   "description": "The VA design system",
   "keywords": [
     "va",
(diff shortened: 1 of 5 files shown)


jaredcunha-usds:
Pushed IE11 fix. Issue was that IE11 has difficulty with the `flex` shortand property, so just had to add to refactor the Sass maps slightly.

packages/formation/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation",
-  "version": "3.2.0",
+  "version": "3.3.0",
```
rianfowler: Not sure why this is showing a change- [master](https://github.com/department-of-veterans-affairs/design-system/blob/master/packages/formation/package.json) is `3.3.0`.  May need to be incremented to `3.4.0`

PR #235, opened 2019-02-01
Gatsby docs

accbjt:
Here is the initial Gatsby project prototype.


.eslintrc
```diff
@@ -90,6 +90,7 @@
     'global-require': 0,
     'no-lonely-if': 0,
     'lines-around-directive': 0,
+    "jsx-a11y/href-no-hash": 0,
   },
```
accbjt: I had to add this because it is causing errors on the Gatsby project. I'm not sure how important this rule is. Let me know if we should remove this.

package.json
```diff
@@ -19,7 +19,9 @@
     "test-debug": "lerna run test-debug --scope=@department-of-veterans-affairs/formation-react",
     "test-watch": "lerna run test-watch --scope=@department-of-veterans-affairs/formation-react"
   },
-  "dependencies": {},
+  "dependencies": {
+    "classnames": "^2.2.6"
+  },
```
accbjt: I was getting a linting error regarding this in formation-react

packages/gatsby-documentation/gatsby-node.js
```diff
@@ -0,0 +1,102 @@
...
+      })
+    }
+  }
+}
+
+exports.createPages = ({ graphql, actions }) => {
+  const { createPage } = actions
+  return new Promise((resolve, reject) => {
+    resolve(
+      graphql(
```
rianfowler: what does Gatsby use GraphQL for? does the build process start a GraphQL server?

packages/gatsby-documentation/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "gatsby-starter-default",
+  "private": true,
+  "description": "A simple starter to get up and developing quickly with Gatsby",
+  "version": "0.1.0",
+  "author": "Kyle Mathews <mathews.kyle@gmail.com>",
```
rianfowler: we can probably clean this up by removing the author and renaming it.

packages/gatsby-documentation/src/components/seo.js
```diff
@@ -0,0 +1,94 @@
...
+              {
+                property: 'og:description',
+                content: metaDescription,
+              },
+              {
+                property: 'og:type',
+                content: 'website',
+              },
+              {
+                name: 'twitter:card',
```
rianfowler: Do we know what this is doing? Probably don't want to embed anything from social media on these pages.

packages/gatsby-documentation/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Project dependencies
+.cache
+node_modules
+yarn-error.log
+
+# Build directory
+/public
```
jbalboni: I don't think this is correctly excluding the public folder. Those files look like they're in this PR.

packages/gatsby-documentation/gatsby-node.js
accbjt: @rianfowler Gatsby is a static generator and it uses GraphQL for querying data for pages. This happens during the build process.

packages/gatsby-documentation/package.json
accbjt: Wow, I missed that thanks

packages/gatsby-documentation/src/components/seo.js
accbjt: This is suppose to be SEO stuff. I left it in because I wasn't sure if we would need it or not. @jbalboni we probably don't need this right.

packages/gatsby-documentation/.gitignore
accbjt: Yeah I don't think so either. In my editor it shows up as ignored but I guess it doesn't. I will look into this more and see why it's not working. I will transfer it into the design-system gitignore file if it doesn't want to use this file. I might have to clear my git cache and see if it removes these files.

.eslintrc
jbalboni: I think you can make an .eslintrc file in the gatsby-docs folder with just this rule and it'll only apply there.

package.json
jbalboni: I think that might have been from not installing dependencies in those projects. We shouldn't have any dependencies in this package.json. Can you remove this, run `npm run bootstrap` and see if you still get the linting error?

packages/gatsby-documentation/LICENSE
```diff
@@ -0,0 +1,22 @@
+The MIT License (MIT)
```
jbalboni: Can you remove this file and update the license in the package.json to match formation and formation-react?

packages/gatsby-documentation/gatsby-browser.js
```diff
@@ -0,0 +1,9 @@
+/* eslint-disable */
```
jbalboni: Looks like we can delete this file

packages/gatsby-documentation/gatsby-ssr.js
```diff
@@ -0,0 +1,7 @@
+/**
```
jbalboni: Delete this, too

packages/gatsby-documentation/src/components/seo.js
rianfowler: The title and meta stuff might be used by the local search component. We can probably ditch the twitter stuff.

jbalboni:
@accbjt This looks good to me aside from the one eslint change and the failing build. Once those are fixed, I think we can merge this.

jbalboni:
@accbjt One other thing I should we should change: can we change the module folder from gatsby-documentation to just documentation? Putting gatsby in the name isn't adding much.

accbjt:
I agree.

accbjt:
@rianfowler  Sure I can.

PR #247, opened 2019-02-13
Border utilty

jaredcunha-usds:
This PR will create a border utility. For documentation on how this utility works, see here: 
https://department-of-veterans-affairs.github.io/vets-design-system-documentation/utilities/border.html#border-shorthand

### DIFF
--- a/packages/formation/sass/utilities/_border.scss
+++ b/packages/formation/sass/utilities/_border.scss
@@ -1,0 +1,322 @@
+$border_sizes: (
+  0,
+  1px,
+  2px,
+  3px,
+  4px,
+  5px,
+  7px,
+  10px
+);
+
+// Standard borders
+@each $size in $border_sizes {
+  .vads-u-border--#{$size} {
+    border: #{$size} solid !important;
+  }
+
+  .vads-u-border-top--#{$size} {
+    border-top: #{$size} solid !important;
+  }
+
+  .vads-u-border-right--#{$size} {
+    border-right: #{$size} solid !important;
+  }
+
+  .vads-u-border-bottom--#{$size} {
+    border-bottom: #{$size} solid !important;
+  }
+
+  .vads-u-border-left--#{$size} {
+    border-left: #{$size} solid !important;
+  }
+}
+
+@each $bp_name, $bp_value in $breakpoints {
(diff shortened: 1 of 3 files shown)


packages/formation/sass/utilities/_border.scss
```diff
@@ -0,0 +1,322 @@
...
+
+.vads-u-border-color--cool-blue-lightest {
+  border-color: $color-cool-blue-lightest !important;
+}
+
+/*
+  Hub colors
+*/
+
+.vads-u-border-color--hub-health-care {
```
jbalboni: I think we'll want to refactor these at some point. We have some generated styles around the hub colors, but I think they're still in vets-website. Would be nice to get this in here and not have extra places to change if a hub color changes.

packages/formation/sass/utilities/_border.scss
```diff
@@ -0,0 +1,322 @@
+$border_sizes: (
+  0,
+  1px,
+  2px,
+  3px,
+  4px,
+  5px,
+  7px,
+  10px
```
jbalboni: Why'd you decide on 10 as the upper limit? Seems like we could go lower, I don't think we have many 10px borders.

jaredcunha-usds: We do have a pseudo element that is 10px wide that resembles a left border on the alerts. That is the primary reason I included that size… should there be any future designs that may riff with that.

jaredcunha-usds: I worked with the design team on to find what other border-widths we are using on the site for the rest of values.

packages/formation/sass/utilities/_border.scss
jaredcunha-usds: Yeah, those variables I think are set in the variables file in `design-system`, but where I can in `vets-website` is they are using the same Sass variables: https://github.com/department-of-veterans-affairs/vets-website/blob/a963d5c02b109b994efbfff35bc3e1758d68821a/src/platform/site-wide/sass/iconography.scss#L64

The variable names were pulled from some USWDS 2.0 palette colors. It might be more ideal to set the colors to more obvious variable names, even if that's declaring them as `$color-hub-health-care: $indigo-cool-60;`.

PR #250, opened 2019-02-18
Updates and fixes for alert modals

U-DON:
These changes will support alert-style modals (PR: department-of-veterans-affairs/vets-website#9596) as described in: department-of-veterans-affairs/vets.gov-team#16806.

Before these changes, passing a `status` prop to any `Modal` instances would result in a modal that appeared to be broken. The logic around handling different status types for modals was outdated especially after the redesign of the `AlertBox` component.

### DIFF
--- a/packages/formation-react/package.json
+++ b/packages/formation-react/package.json
@@ -34,7 +34,7 @@
     "window-or-global": "^1.0.1"
   },
   "peerDependencies": {
-    "@department-of-veterans-affairs/formation": "^2.0.0",
+    "@department-of-veterans-affairs/formation": "^4.0.0",
     "react": "^15.5.4||^16.7.0"
   }
 }
--- a/packages/formation-react/package.json
+++ b/packages/formation-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation-react",
-  "version": "3.1.1",
+  "version": "4.0.0",
   "description": "React implementation of patterns in Formation",
   "keywords": [
     "react",
--- a/packages/formation-react/src/components/modal/Modal.jsx
+++ b/packages/formation-react/src/components/modal/Modal.jsx
@@ -159,7 +159,6 @@ Modal.propTypes = {
   cssClass: PropTypes.string,
   /**
    * Id of the modal, used for aria attributes.
-   * It will be auto-generated if not provided.
    */
   id: PropTypes.string,
   /**
--- a/packages/formation-react/src/components/modal/Modal.jsx
+++ b/packages/formation-react/src/components/modal/Modal.jsx
@@ -1,21 +1,19 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import classNames from 'classnames';
+import _ from 'lodash';
 
 const ESCAPE_KEY = 27;
 
 class Modal extends React.Component {
   constructor(props) {
     super(props);
-    this.state = {
-      lastFocus: null,
-    };
+    this.state = { lastFocus: null };
   }
 
   componentDidMount() {
-    if (this.props.visible) {
-      this.setupModal();
-    }
+    this.id = this.props.id || _.uniqueId('va-modal-');
+    if (this.props.visible) this.setupModal();
   }
 
   componentDidUpdate(prevProps) {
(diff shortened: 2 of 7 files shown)


packages/formation/sass/modules/_m-modal.scss
```diff
@@ -22,97 +21,19 @@
...
-  }
-
   .va-modal-body {
     p:first-of-type {
       margin-top: 0;
     }
   }
 
-  // "Forked" modal styles
-  &-split {
```
U-DON: Saw no mention of a `va-modal-split` anywhere in our code (vets-website, vagov-content or this repo).

packages/formation/sass/modules/_m-modal.scss
```diff
@@ -164,6 +85,68 @@
       }
     }
   }
+
+  &-alert {
+    .va-modal-inner {
+      max-width: 60rem;
```
U-DON: Didn't want to affect all modals by changing the original `max-width: 40rem;`.

jbalboni:
FYI @jaredcunha-usds

packages/formation-react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation-react",
-  "version": "3.1.1",
+  "version": "4.0.0",
```
jbalboni: Can you also update the peer dep version for `formation` in here?

packages/formation-react/src/components/modal/Modal.jsx
```diff
@@ -1,21 +1,19 @@
...
-      lastFocus: null,
-    };
+    this.state = { lastFocus: null };
   }
 
   componentDidMount() {
-    if (this.props.visible) {
-      this.setupModal();
-    }
+    this.id = this.props.id || _.uniqueId('va-modal-');
```
jbalboni: What prompted you to add the auto generated id? I know we do that in other components, but it'd be nice to get away from that, I think. Keeps us from needing to rely on Lodash in lots of components.

packages/formation/sass/modules/_m-modal.scss
```diff
@@ -164,6 +85,68 @@
...
+    }
+
+    .va-modal-body {
```
jbalboni: Would be interesting to see how it would look if we refactored this component to use the new utilities. I think we'd avoid all this nesting, but the component source code would be more complicated. cc @jaredcunha-usds 

(Not saying we should do that now, just for the future)

packages/formation-react/src/components/modal/Modal.jsx
U-DON: Yeah, I'd like to not use Lodash here, but it looks like we use `id` for ARIA labelling purposes, but a lot of modals are not passing any ids. I guess they will just end up with headers that `#undefined-title`, which can still work. On the other hand, sometimes modals don't even pass any titles, and there won't be any title corresponding to the modal's `aria-labelledby`. Not sure of the best way to resolve all of those ARIA issues.

U-DON: I suppose I can just give the title a generic id of `#modal-title` and conditionally set the `aria-labelledby` if there's a title. I don't think we have any scenarios where we can have multiple modals up at the same time, since any hidden ones will just be `<div />`s.

jbalboni: Yeah, I think that would work

packages/formation-react/src/components/modal/Modal.jsx
```diff
@@ -86,27 +84,30 @@ class Modal extends React.Component {
...
+        {primaryButton && <button className="usa-button" onClick={primaryButton.action}>{primaryButton.text}</button>}
+        {secondaryButton && <button className="usa-button-secondary" onClick={secondaryButton.action}>{secondaryButton.text}</button>}
       </div>
     );
+  };
+
+  render() {
+    const { status, title, visible } = this.props;
 
     if (!visible) { return <div/>; }
```
U-DON: Going to change this to return `null`. Was worried about breaking some tests, but it'll be cleaner markup to not render anything if it's supposed to be hidden. Also the recommendation from React: https://reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering

U-DON:
It's come to my attention that there is a plan to do a site-wide redesign of modal close buttons as depicted in the mocks of the related issue. So I went ahead and updated the modal close button as well.

> screen shot 2019-02-19 at 2 07 40 am

packages/formation/sass/modules/_m-modal.scss
jaredcunha-usds: Yeah - I think this would be a good use of utilities. You can use a an alert component (current naming convention or create new `vads-c-alert`), then you can use the utilities to provide options for structuring content within them.

U-DON:
Looks like the background is supposed to be white even for alert modals, so updated:

screen shot 2019-02-19 at 12 03 48 pm

PR #251, opened 2019-02-19
Remove foundation visibility

accbjt:
Removing visibility classes for this ticket

https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/vets.gov-team/5516

### DIFF
--- a/packages/formation/package.json
+++ b/packages/formation/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation",
-  "version": "3.14.1",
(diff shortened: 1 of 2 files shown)


packages/formation/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/formation",
-  "version": "3.14.0",
```
jbalboni: This is technically a breaking change, so I think we need to bump this up to major version 4.

